PR #1, opened 2019-01-07
Changed repo structure

Mikerah:
The directory structure is outdated. I have changed it to make it much simpler and easier to add to in the future. Initially, I will be coding up all gossip sub functionality in index.js. If needed, we can later refactor into more files and changes the repo structure.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -425,8 +425,13 @@
         "pem-jwk": "^1.5.1",
         "protons": "^1.0.1",
         "rsa-pem-to-jwk": "^1.1.3",
-        "tweetnacl": "^1.0.0",
-        "webcrypto-shim": "github:dignifiedquire/webcrypto-shim#190bc9ec341375df6025b17ae12ddb2428ea49c8"
+        "tweetnacl": "^1.0.0"
+      },
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


GregTheGreek:
Do we want to do it in typescript for consistency?

Mikerah:
I would say no because the eventually go I had in mind was to integrate this into [js-libp2p](https://github.com/libp2p/js-libp2p). From what I know about libp2p's roadmap for this year, they are focusing on building new communities around other implementations of libp2p (it doesn't seem to include Typescript).

It does seem like there are others that want to have some form of Typescript compatibility. For example, this [issue](https://github.com/libp2p/js-libp2p/issues/208) was added to their backlog but has no activity.

I would suggest that we initially do a JS one as a placeholder because this does need to be worked on. Later one, we can pick up where that issue left off and contribute even more to libp2p.

Thoughts?

GregTheGreek:
> I would say no because the eventually go I had in mind was to integrate this into js-libp2p.

Wth that being said, i agree.

> I would suggest that we initially do a JS one as a placeholder because this does need to be worked on. Later one, we can pick up where that issue left off and contribute even more to libp2p.

This is probably the best approach.

PR #15, opened 2019-02-06
Finished an initial version of GossipSub

Mikerah:
I completed an initial version of gossipsub (I think). The code is based on this [specification](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) and the [reference implementation in Go](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go). The code most likely can't run due to several bugs that I have eyeballed and some missing functionality in other libp2p modules that I plan on working on. 

I have started writing tests and I will be submitting PRs for these tests as I go along. In the meantime, any extra cleanup and bug fixes will be greatly appreciated.

Current blockers:
- Publish doesn't work. In the 2 node tests, the node that receives the message from a topic that it is subscribed to does receive a graft message but it doesn't send a I want message. As a result, it doesn't get the actual message.


raulk:
@Mikerah – would you like @vasco-santos @jacobheun to have a peek as well?

Mikerah:
@raulk If they have the time, that would be great!

src/index.js
```diff
@@ -0,0 +1,756 @@
...
+
+       let out = _rpcWithControl(null, null, null, null, prune)
+       if(peer && peer.isWritable()) {
+          peer.write(out)
+	  peer.sendUnsubscriptions([topic])
+       }
+
+   }
+
+   _heartbeatTimer() {
```
wemeetagain: This function needs cleanup, appears to have duplicate code, currently causing errors trying to create a new GossipSub object.

src/message/index.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+const protons = require('protons')
```
wemeetagain: Add `protons` as a dependency

src/index.js
Mikerah: Currently, this function doesn't even work. It needs serious debugging. I plan on working on this throughout the week while I write tests.

index.js
```diff
@@ -0,0 +1,228 @@
+'use strict'
```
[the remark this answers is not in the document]
Mikerah: I have updated the repo structure. Everything should be in `/src` and some should be remove due to the dependency on `js-libp2p-pubsub`.

src/index.js
wemeetagain: :+1:

src/message/topic-descriptor.proto.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+module.exports = `
+// topicCID = cid(merkledag_protobuf(topicDescriptor)); (not the topic.name)
```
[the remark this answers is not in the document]
Mikerah: I copied this file from `js-libp2p-floodsub`. I'm not sure what purpose of that line is. Would it be safe to remove it all together since it has nothing to do with the topic descriptor protocol buffer?

src/messageCache.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Adds a message to the current window and the cache
+     *
+     * @param {pb.rpc.RPC.Message Object}
+     *
+     */
+    put (msg) {
+	var msgID = utils.msgId(msg.from, msg.seqno)
```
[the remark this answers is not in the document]
Mikerah: I inconsistently used `var` in this file only. I tend to slip into poor habits when learning something difficult. I will be more consistent in the future.

As for your second point, I'm not quite sure I understand. I copied over the `utils.js` from `js-libp2p-floodsub`. The definition of `utils.msgId` is such that you need to pass in both `msg.from` and `msg.seqno`. This can very easily be amended later on.

src/message/topic-descriptor.proto.js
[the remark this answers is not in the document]
vasco-santos: let's remove it then in both projects!

src/messageCache.js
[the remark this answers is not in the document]
vasco-santos: No problem, just for making sure everything gets great!

We can probably move the utils to `js-libp2p-pubsub` and refactor it to my suggestion. What do you think? Then, both floodsub and gossipsub require the the common utils from there

src/messageCache.js
```diff
@@ -0,0 +1,115 @@
...
+    constructor (gossip, history) {
+	/**
+	 * @type {Map<string, RPC.Message>}
+	 */
+        this.msgs = new Map()
+
+	/**
+	 * @type {Array<Array{CacheEntry}>}
+	 */
+	this.history = []
```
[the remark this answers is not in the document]
Mikerah: Oddly, making this change breaks some of the tests I have written. I have verified that both methods for populating the multidimensional `this.history` array are equivalent. I have tried searching for why this might be but found conflicting reasons. I will keep it as it is for the moment.

GregTheGreek:
@Mikerah make sure to resolve your conversations on github (above comments) if you've fixed them in a commit :)

src/index.js
```diff
@@ -0,0 +1,756 @@
...
+const GossipSubHistoryGossip = 3
+
+// Heartbeat interval
+const GossipSubHeartbeatInitialDelay = 100 // In milliseconds
+const GossipSubHeartbeatInterval = 1 // In seconds
+
+// Fanout ttl
+const GossipSubFanoutTTL = 60 // in seconds
+
+
```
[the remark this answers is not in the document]
Mikerah: The description in the link you posted pertains to the fanout mapping of mesh peers. The description that you refer to is use when checking if a topic hasn't been published to in a while. I don't see what you mean?

src/index.js
```diff
@@ -0,0 +1,756 @@
+'use strict'
+
```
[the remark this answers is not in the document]
Mikerah: I think it makes more sense to have a link to the spec in the readme instead.

src/index.js
```diff
@@ -0,0 +1,756 @@
...
+	   return
+	}
+
+	iwantMsgs.forEach(function(iwantMsg) {
+	    let iwantMsgIDs = iwantMsg.MessageIDs
+	    if(!(iwantMsgIDs || iwantMsgIDs.length)) {
+	        return 
+	    }
+
+            iwantMsgIDs.forEach(function(msgID){
```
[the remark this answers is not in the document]
Mikerah: This is a neat trick. However, I think it sacrifices code readability in favor of efficiency.

src/index.js
```diff
@@ -0,0 +1,756 @@
...
+    }
+    
+    /**
+     * Subscribes to a topic
+     * @param {String}
+     *
+     */
+   subscribe(topic) {
+       assert(this.started, 'GossipSub has not started')
+       if (this.mesh.has(topic)) {
```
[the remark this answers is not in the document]
Mikerah: I guess this could be written as an assertion instead of an if. It could be omitted completely as well.

GregTheGreek:
@vasco-santos @Mikerah I'm going to do a spec review right now, hopefully that can help speed a few things up :)

src/index.js
```diff
@@ -0,0 +1,849 @@
...
+	   peers.forEach((peer) => {
+	       if(this.topics.has(peer)) {
+	           peers.delete(peer)
+	       }
+	   })
+
+	   // do we need more peers?
+	   if (peers.size < constants.GossipSubD) {
+	       let ineed = constants.GossipSubD - peers.size
+               peersSet = this._getPeers(topic, ineed)
```
GregTheGreek: Needs a `let` statement

src/index.js
```diff
@@ -0,0 +1,849 @@
...
+
+   }
+
+   _emitGossip (topic, peers) {
+       let messageIDs = this.messageCache.getGossipIDs(topic)
+       if(!messageIDs.length) {
+           return
+       }
+
+       gossipSubPeers = this._getPeers(topic, constants.GossipSubD)
```
GregTheGreek: needs a `let` statement

src/index.js
```diff
@@ -0,0 +1,849 @@
...
+     * @override
+     * @param {Peer} peer
+     * @returns {undefined}
+     */
+    _removePeer (peer) {
+        const id = peer.info.id.toB58String()
+	
+	this.log('remove', id, peer._references)
+	// Only delete when no one else if referencing this peer.
+	if (--peer._references === 0){
```
GregTheGreek: the `--` to me seems that if `peer._references === 2` then this condition is met. It seems to me that you are actually dropping the peer if they have `2` references

src/index.js
```diff
@@ -0,0 +1,849 @@
...
+	    this.log('delete peer', id)
+            this.peers.delete(id)
+	
+            // Remove this peer from the mesh
+            for(let [topic, peers] of this.mesh.entries()){
+	        peers.delete(peer)
+	    }
+	    // Remove this peer from the fanout
+            for (let [topic, peers] of this.fanout.entries()){
```
GregTheGreek: since you're not using `topic` both here and on `line 89` I would change it to and underscore to identify you're not using that. `topic` -> `_`

package.json
```diff
@@ -20,7 +20,15 @@
...
     "libp2p": "^0.23.1",
-    "libp2p-floodsub": "^0.15.0"
+    "libp2p-floodsub": "^0.15.7",
+    "libp2p-pubsub": "0.0.1",
+    "protons": "^1.0.1",
+    "time-cache": "^0.3.0"
+  },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "mocha": "^5.2.0"
```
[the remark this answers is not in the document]
Mikerah: There's a PR waiting to be reviewed that adds aegir to this repo.

src/index.js
Mikerah: `--` is the decrement operator. If `peer._references === 2` then `--peer._references === 1`.

GregTheGreek:
@Mikerah instead of doing a merge commit you should consider doing a rebase `git rebase origin master` in your branch which should clean things up

src/index.js
GregTheGreek: Sorry yes, I meant to use `1` not `2`.

None the less, why are you performing this operation? Is it assuming that you are the only peer? i.e if `_references === 1` you are the `1` therefore `drop`?

Mikerah: if `peer._references == 1` then there is 1 other peer (not equal to `peer`) referencing `peer` and so `peer` should not be dropped.

GregTheGreek: You've lost me, assumptions:
- `this` means the current user
- `peer` is another client 

`peer._references === 1`  means that `this` is the `1` reference therefore because `this` is the only one connecting to `peer` we drop them?

Mikerah:
I have updated the PR description to include some of the current blockers.

raulk:
@Mikerah Hey! Just wanted to check in with you to see if you needed any assistance, review, help, guidance, etc. from us to land this great piece of work.

CC @vasco-santos

Mikerah:
Hi @raulk,

I have been debugging for the past several weeks. I'm just starting to see some progress and I hope to finish testing the 2 gossipsub nodes case soon. It would be great to have another review once that is finished. Afterwards, I plan on working on the multiple node case.

src/index.js
```diff
@@ -0,0 +1,864 @@
...
+
+        if (msgs && msgs.length) {
+            this._processRpcMessages(utils.normalizeInRpcMessage(msgs))
+        }
+
+        if (!controlMsg) {
+            return
+        }
+
+        let iWant = this._handleIHave(idB58Str, controlMsg)
```
wemeetagain: It might make sense to pass in the array of each control message type here.
eg: `this._handleIHave(peer, controlMsg.ihave)`

src/index.js
```diff
@@ -0,0 +1,864 @@
...
+     * @param {RPC.control} controlRpc
+     *
+     * @returns {Array<RPC.Message>}
+     */
+    _handleIWant (peer, controlRpc) {
+        // @type {Map<string, RPC.Message>}
+        let ihave = new Map()
+
+        let iwantMsgs = controlRpc.iwant
+        if (!iwantMsgs){
```
wemeetagain: This comment is pretty subjective so feel free to ignore.
There are some cases of null checks and short-circuiting that I think are somewhat unnecessary if we assume that they are internal methods. Imo makes things trickier than need be.

In this case, if `controlRpc` is indeed a control object, it will always have an array `controlRpc.iwant` and any elements of the array will always have a property `messageIDs`. We would want to input guard for methods exposed to users, but if this is an internal method, we can do without the extra checks.

You might also consider always returning the same type of object in the short circuits, so you don't change the function return type (to `Array<RPC.Message> | undefined`). This will make processing the outputs less error-prone since the return type will always be the same.

src/index.js
```diff
@@ -0,0 +1,864 @@
...
+        let iHave = this._handleIWant(idB58Str, controlMsg)
+        let prune = this._handleGraft(idB58Str, controlMsg)
+        this._handlePrune(idB58Str, controlMsg)
+
+        if(!(iWant || iWant.length) && !(iHave || iHave.length) && !(prune || prune.length)) {
+            return
+        }
+
+
+        let outRpc = this._rpcWithControl(iHave, null, iWant, null, prune);
```
wemeetagain: needs a `msgs` argument passed in.

vasco-santos:
@Mikerah @raulk 

Ping me when you are ready. I want to review this side by side with the spec once you feel this ready. Any other ways we would like help, just reach out

src/index.js
```diff
@@ -0,0 +1,864 @@
...
+     * @returns {void}
+     *
+     */
+    _processConnection (idB58Str, conn, peer) {
+        pull(
+            conn,
+            lp.decode(),
+            pull.map((data) => RPC.decode(data)),
+            pull.drain(
+                (rpc) => this._onRpc(idB58Str, rpc),
```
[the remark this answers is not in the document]
Mikerah: So, I'm planning on turning this part of the code into an async iterator, anyway. I can take this into account when doing that.

src/index.js
```diff
@@ -0,0 +1,903 @@
...
+const Pubsub = require('libp2p-pubsub')
+const pull = require('pull-stream')
+const lp = require('pull-length-prefixed')
+const nextTick = require('async/nextTick')
+
+const MessageCache = require('./messageCache').MessageCache
+const utils = require('./utils')
+const assert = require('assert')
+
+const RPC = require('./message').rpc
```
vasco-santos: The way this `require` is being made + how it is used is causing `subscribe` and`publish` to fail.

```suggestion
const { rpc } = require('./message')
```

Then for using this, you should use `rpc.RPC.encode()` and `rpc.RPC.decode()`

Basically, when a peer receives a message from `subscribe`, the `_processConnection` is failing because of the current `RPC.encode()` function does not exist. Also, the messages sent like the `_sendGraft` should also have the extra property `RPC`. 

If you prefer, you can also do the require in a way that you don't need to change the rest of the code.

src/utils.js
```diff
@@ -0,0 +1,95 @@
+'use strict'
```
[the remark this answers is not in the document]
Mikerah: I just submitted a PR to `libp2p-pubsub`.

test/utils.spec.js
```diff
@@ -0,0 +1,78 @@
...
+    const first = utils.randomSeqno()
+    const second = utils.randomSeqno()
+
+    expect(first).to.have.length(20)
+    expect(second).to.have.length(20)
+    expect(first).to.not.eql(second)
+  })
+
+  it('msgId', () => {
+    expect(utils.msgId('hello', Buffer.from('world'))).to.be.eql('hello776f726c64')
```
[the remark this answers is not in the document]
Mikerah: This was copied over from `js-libp2p-floodsub`. I think this test should be moved to `js-libp2p-pubsub` since we are also moving `utils.js` over there.

wemeetagain:
@vasco-santos ready for another glance, when you have time.

We're passing the tests at least. The tests currently only test pubsub-oriented behavior since they originated from the floodsub tests, and they don't do a good job at testing the gossipy things (eg: adding/removing mesh peers on heartbeat, maintaining the gossip message cache, maintaining the fanout peers, etc). I plan on writing a few more tests to exercise the code from a gossipsub perspective.

src/index.js
```diff
@@ -0,0 +1,979 @@
...
+   * Send graft and prune messages
+   *
+   * @param {Map<Peer, Array<String>>} tograft
+   * @param {Map<Peer, Array<String>>} toprune
+   */
+  _sendGraftPrune (tograft, toprune) {
+    for (const [p, topics] of tograft) {
+      const graft = topics.map((topicID) => ({ topicID }))
+      let prune = null
+      if (toprune.has(p)) {
```
vasco-santos: get before

test/2-nodes.js
```diff
@@ -0,0 +1,544 @@
...
+const promisify = require('promisify-es6')
+
+const GossipSub = require('../src')
+const utils = require('./utils')
+const first = utils.first
+const createNode = utils.createNode
+const expectSet = utils.expectSet
+
+describe('basics between 2 nodes', () => {
+  describe('fresh nodes', () => {
```
vasco-santos: We should be able to run every single test in this suite of tests individually, which is not the case because we have "tests" for:

- `mount the pubsub protocol`
-  `start both GossipSubs`
- `Dial from nodeA to nodeB`

Those tests are essential for the remaining ones. For instance, if I do a `it.only(...)` for one of the following tests, it will fail because it will need the previous ones to run.

What I recommend here is to have multiple befores for all the code that must run for each test:

```js
// Create nodes
before(async () => 
....

// mount the pubsub protocol
before(....

// start both GossipSubs
before(...

// Dial from nodeA to nodeB
before(...
```

However, there are other dependencies here between the remaining tests. In this case, you can also use `beforeEach` and `afterEach`, which will run before and after each test, in order to setup specific behavior for each of them.

There are other approaches as well, I am ok with other solutions but the tests must be completely isolated. If you need assistance on this, just ping me on IRC

src/index.js
```diff
@@ -0,0 +1,979 @@
...
+        }
+      })
+      this.emit('meshsub:subscription-change', peer.info, peer.topics, subs)
+    }
+
+    if (msgs.length) {
+      this._processRpcMessages(utils.normalizeInRpcMessages(msgs))
+    }
+
+    if (!controlMsg) {
```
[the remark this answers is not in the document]
wemeetagain: I think we want to still process the `msgs` and `subscriptions` even if there are no `control` messages.

Eg: above, L193, the `_processRpcMessages` is where we forward messages on to other peers. 
We want to forward messages regardless of if there is eg: a graft/iwant/etc control message.
Likewise, if a peer sends a `SUBSCRIBE` but no control messages, we still want to process the subscriptions.

src/index.js
```diff
@@ -0,0 +1,979 @@
...
+   *
+   * @param {Peer} peer
+   * @param {rpc.RPC.controlMessage Object} controlRpc
+   *
+   * @returns {rpc.RPC.ControlIWant Object}
+   */
+  _handleIHave (peer, controlRpc) {
+    let iwant = new Set()
+
+    let ihaveMsgs = controlRpc.ihave
```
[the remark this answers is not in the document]
wemeetagain: Looking at the protobuf definition, I think `ihave` will never be `null`, only the `RPC.control` parent object.
If `RPC.control` is not `null`, it will have `ihave`, `iwant`, `graft`, and `prune` arrays.

src/index.js
```diff
@@ -0,0 +1,979 @@
...
+      // Send GRAFT to mesh peers
+      if (this.fanout.has(topic)) {
+        this.mesh.set(topic, this.fanout.get(topic))
+        this.fanout.delete(topic)
+        this.lastpub.delete(topic)
+      } else {
+        const peers = this._getPeers(topic, constants.GossipSubD)
+        this.mesh.set(topic, peers)
+      }
+      this.mesh.get(topic).forEach((peer) => {
```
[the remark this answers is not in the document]
wemeetagain: I think in either case, `mesh[topic] = gmap`, and then `gmap` is iterated through.

If `fanout[topic]` exists, we use those peers, set `mesh[topic]`, then delete `fanout[topic]`.
In the other case, we just find some gossip peers in the topic and set `mesh[topic]`.

src/index.js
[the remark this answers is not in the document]
vasco-santos: Yes, thanks for making this clear, you are right!

src/index.js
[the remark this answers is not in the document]
vasco-santos: Correct, I misunderstood the set!

src/index.js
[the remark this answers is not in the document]
vasco-santos: The point here is that you don't need to do the next validation if you add this fallback, which IMHO is way easier for readability.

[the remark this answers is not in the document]
wemeetagain: Gotcha. Agreed, much more readable assuming its an Array.

test/2-nodes.js
wemeetagain: Yeah, these tests are not the best. They are inherited from `js-libp2p-floodsub`, and that repo has the same issues you mention here.

I'm thinking it may almost be easier to scrap the existing tests and redo them from scratch, testing and setting up each item independently.
They also don't test the gossipsub components at all (I noticed several bugs in the rpc handling while addressing your review).

I'll work on revamping the testing to be more complete and flexible.

wemeetagain:
@vasco-santos Thanks for the review! :smiley_cat: 
I think I've addressed all of your comments except testing. Happy to fix anything else you think needs fixing, too.

I'm planning a deeper look at testing today, will ping you here when its ready.

test/2-nodes.js
vasco-santos: yes! `libp2p-floodsub` tests are terrible and definitely not a good example! Unfortunately, I got the responsibility for the repo in that state, and I did not have the bandwidth to refactor them.

vasco-santos:
Happy to help @wemeetagain 

Another thing that I have been thinking that it would be worth to refactor here is the `heartbeat` logic. The `src/index.js` file is massive and we can have it way more modular. 

The first obious candidate would be the `heartbeat`. We can have something similar to the codebase we have in the dht for the random-walk (https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/random-walk.js). That is, a class for `heartbeat`, with start and stop, and the periodic logic. It should be an easy refactor that will decrease the complexity of the remaining code. There are other things that could be changed to another place, as finding stuff in a file with almost 1000 lines like `src/index.js` is too difficult.

wemeetagain:
@vasco-santos 
I cleaned up the tests to run independent of one another, but haven't done extensive testing on the gossip-specific functionality.

For refactoring, I was able to pull out basic pubsub functionality from the gossip/mesh functionality. Perhaps this can be merged upstream into libp2p-pubsub.

I'll keep working at expanding testing and other refactoring experiments.

vasco-santos:
@wemeetagain @Mikerah 

FYI, we need to do a change here regarding signing and verification of messages, as this was changed recently in `go-libp2p` and `js-libp2p-floodsub`. More context on [libp2p/js-libp2p-floodsub#78](https://github.com/libp2p/js-libp2p-floodsub/pull/78). This will be abstracted in the future in `js-libp2p-pubsub`, but for not adding too much overhead now to guarantee interop, we decided to start by doing this individually.

.travis.yml
```diff
@@ -26,17 +26,5 @@ jobs:
         - npx aegir dep-check
```
vasco-santos: Can you remove the following line?

```js
npx aegir commitlint --travis
```

This is making the CI to fail. We have been discussing this, and we can remove that as this seems to be a difficulty to contributors

src/index.js
```diff
@@ -0,0 +1,725 @@
...
+  _piggybackGossip (peer, outRpc, ihave) {
+    outRpc.control.ihave = ihave
+  }
+
+  /**
+   * Maintains the mesh and fanout maps in gossipsub.
+   *
+   * @returns {void}
+   */
+  _heartbeat () {
```
vasco-santos: I would like to have the `heartbeat` logic in a separate file, a class that receives a reference to the properties it needs from here, and has the start and stop functions, which are called from here.

wemeetagain:
@vasco-santos 
I pulled out the heartbeat and added message signing.

Do you have an idea of what we'd want to be testing? Here's some things I was thinking:
- test heartbeat occuring with proper regularity
- test adding peers into topic beyond mesh limit, ensure limit is set
- test gossip is being sent to the right peers

other thoughts?

Mikerah:
@wemeetagain Test piggybacking. I know this is implicit but just for extra measure, as it is more of an optional feature.

vasco-santos:
We should also test a network with some nodes running floodsub and others gossipsub, and guarantee the expected behavior. 

Other than that, stress tests would be rad, but I think it may come in a next PR, while we are also working on getting it into `js-libp2p` and creating the `interop` tests.

wemeetagain:
@vasco-santos 
Added all of the above tests minus the floodsub interop test.
In working through that test, I believe that floodsub interop functionality will require some sort of (potentially minor) change to pubsub.
Specifically, even if we add `this.libp2p.handle(constants.FloodSubID, this._onConnection)` to the appropriate section of the gossipsub `start` method, the connection to a floodsub peer will error [here](https://github.com/libp2p/js-libp2p-pubsub/blob/master/src/index.js#L157) because of the way pubsub only dials a single protocol [here](https://github.com/libp2p/js-libp2p-pubsub/blob/master/src/index.js#L149).

Mikerah:
@wemeetagain For this test, what needs to be done is to make sure that floodsub peers added to a gossipsub peer's mesh receives a subscription as if it were a floodsub peer, not a gossipsub peer. So, from the floodsub peer's perspective, it cannot treat a gossipsub peer as a floodsub peer so a gossipsub peer has to treat the floodsub peer as it is.

So, instead of handling floodsub peers in the start method, what you can do is once a peer is connected, determine what protocol is it using. It's `peerInfo` object should have `this.protocols` populated with protocols that it supports. If that peer is a floodsub peer, just send a message to it as per the floodsub spec.

wemeetagain:
@Mikerah 
The problem is that a stream is never properly created. See the links in my previous comment.

I'm not talking about handling, in a general sense, the floodsub peers in the `start` method. I'm referring to the `libp2p.handle` method that gets called in the pubsub `start` method, which sets up a connection callback when a peer of a certain protocol is dialed. See [here](https://github.com/libp2p/js-libp2p-pubsub/blob/master/src/index.js#L305)

The logic to maintain subscriptions and send messages appropriately to floodsub peers is already present (though untested).
The "issue" is that the pubsub module assumes the router has only a single protocol, and this is causing floodsub peers to not be properly connected, and this can only be fixed by refactoring pubsub or doing an unreasonable amount of method specialization in gossipsub.

vasco-santos:
@wemeetagain that is indeed a problem! Can you PR `js-libp2p-pubsub` changing the `libp2p.dialProtocol`?

wemeetagain:
@vasco-santos 
How would you feel about merging in our current work and finishing out floodsub/golang/etc interop in other PRs?

src/pubsub.js
```diff
@@ -0,0 +1,368 @@
...
+    // Broadcast UNSUBSCRIBE to all peers
+    this.peers.forEach((peer) => {
+      peer.sendUnsubscriptions(topics)
+    })
+
+    this.leave(unTopics)
+  }
+
+  leave (topics) {
+    // noop - add implementation to subclass
```
vasco-santos: same as above

src/pubsub.js
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    asyncMap(messages, buildMessage, (err, msgObjects) => {
+      if (err) callback(err)
+      this._publish(utils.normalizeOutRpcMessages(msgObjects))
+    })
+  }
+
+  _publish (rpcs) {
+    // noop - add implementation to subclass
```
vasco-santos: same as above

test/gossip.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+  })
+  afterEach(async function () {
+    this.timeout(6000)
+    await Promise.all(nodes.map((n) => stopNode(n.gs)))
+    await Promise.all(nodes.map((n) => stopNode(n)))
+  })
+
+  it('should send gossip to non-mesh peers in topic', async function () {
+    this.timeout(0)
```
vasco-santos: Why this?

src/heartbeat.js
```diff
@@ -0,0 +1,169 @@
+/* eslint-disable valid-jsdoc */
```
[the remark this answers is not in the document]
wemeetagain: Not sure, if I had to guess, this was originally in floodsub. This file was copied from src/index.js (which I believe was inherited from floodsub). I'll try removing.

test/gossip.js
wemeetagain: It took longer than 2 seconds, ~5 seconds. I can try to dial it in so it is at least bounded.

README.md
```diff
@@ -1,3 +1,24 @@
...
+* [API](#API)
+* [Contribute](#Contribute)
+* [License](#License)
+
+## Overview
+Gossipsub is an implementation of pubsub based on meshsub and floodsub. You can read the specification [here](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub).
+
+## Install
+
+## Usage
```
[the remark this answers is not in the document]
wemeetagain: it looks like the API documentation was autogenerated in both cases.
I left the API section blank for now.

PR #26, opened 2019-05-24
chore: update package.json

wemeetagain:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "libp2p-gossipsub",
-  "version": "0.0.0",
-  "description": "An experimental Javascript implementation of gossip sub",
+  "version": "0.1.0",
```
vasco-santos: How do you feel about releasing `0.0.1` first?

I would like to have `0.1.0` with interop tests, but it is not a strong opinion

wemeetagain:
@vasco-santos I see why we reset the version to 0.0.0 initially.
It looks like `aegir release` bumps the version automatically.

vasco-santos:
Oh yes, `aegir release` bumps the version automatically. Once you release the initial version to npm please ping me

PR #29, opened 2019-05-29
chore: enable all browser tests

vasco-santos:
All Browser tests enabled in this module

Currently, there is a flaky test in this PR that needs to be addressed.

@wemeetagain @Mikerah any idea why this is happening? It seems that sometimes we are trying to dial with a problematic peer.


GregTheGreek:
Any update on this?

vasco-santos:
@GregTheGreek I rebased this PR and skipped the browser tests which were flaky.

We can go with this PR merged if you want, as it will be better to have some browser tests than none. After that, I can open issues to unskip those tests. 

I don't know what the problem is, but I did not have availability for trying to figure it out too.

GregTheGreek:
@vasco-santos Ok thanks for the update.

@wemeetagain Any thoughts?

PR #30, opened 2019-05-30
[WIP] Basic Benchmarks

Mikerah:
I have started writing some basic benchmarks for gossipsub based on [outdated benchmarks for floodsub](https://github.com/libp2p/js-libp2p-floodsub/tree/master/benchmarks). Ideally, these benchmarks should be refactored into js-libp2p-pubsub and the benchmarks in this repo should be gossipsub specific (benchmarking heartbeats,etc).

For now, this will do and more extensive benchmarking will be done in subsequent weeks.

cc'ing @vasco-santos and @wemeetagain


Mikerah:
By the way, these benchmarks also "hang". So, they still don't work.

vasco-santos:
FYI @Mikerah , regarding pubsub tests, I updated the [libp2p/js-libp2p-floodsub/blob/chore/update-benchmark/benchmarks/index.js](https://github.com/libp2p/js-libp2p-floodsub/blob/chore/update-benchmark/benchmarks/index.js) tests in `floodsub`. Tomorrow I will create a PR with this, but you may be able to get this working by replicating what I changed in `floodsub`.

I think we can go with this basic benchmark for now, and create new ones in subsequent weeks

PR #33, opened 2019-07-08
chore: remove unused deps and move libp2p to a test dep

jacobheun:
A couple of the dependencies listed were not being used. Libp2p was also being included as a dependency instead of a dev dependency, and I also bumped it to the latest release from the RC. 

This should reduce the bundle size.


package.json
```diff
@@ -54,6 +51,7 @@
     "chai-spies": "^1.0.0",
     "detect-node": "^2.0.4",
     "dirty-chai": "^2.0.1",
+    "libp2p": "~0.25.4",
```
mpetrunic: Have you considered putting libp2p as peer dependency so lib can define min libp2p version on which is working?

jacobheun: I think there's definitely value in doing that across the libp2p modules. The usage of gossipsub will ultimately require a new version of libp2p (likely 0.26) as its currently hardcoded to use floodsub, so specifying that here would be helpful. 

If we want to move forward with that here, I think it would be best to create an issue to track adding that once the new version of libp2p is out with support to plug gossipsub instead of floodsub.

PR #34, opened 2019-07-08
feat: fallback to floodsub

vasco-santos:
Added a fallback to dial with other peers running floodsub, behind an option.

For this context, the `_dialPeer` function was override from `libp2p-pubsub` to extend the dial behavior with the selection of the appropriate protocol to dial .


src/pubsub.js
```diff
@@ -50,6 +61,65 @@ class BasicPubSub extends Pubsub {
...
+    this._dials.add(idB58Str)
+
+    this.log('dialing %s', idB58Str)
+
+    // Define multicodec to use
+    let multicodec = this.multicodec
+
+    // Should fallback to floodsub if no Gossipsub and fallback enabled
+    if (this._options.fallbackToFloodsub &&
+      !peerInfo.protocols.has(this.multicodec)) {
```
jacobheun: The issue here is that if we are dialing the peer for the first time we won't know their protocols, so this is going to return false and use floodsub even if gossipsub is actually supported.

I think this logic can be moved inside of `dialProtocol`. The idea being that we can attempt to negotiate gossipsub right away. If the dial fails and we get an `err`, we should be able to see if this is due to a protocol unsupported error. Since the `Identify` protocol runs in parallel, we're not guaranteed to know the protocols yet, so we should inspect the error (we will probably want to update the multistream-select error to provide a code as I dont believe it currently does).

Additionally, if we wanted to try and avoid blind protocol negotiation when possible, we could check to see if the `peerInfo.protocol` set is empty, and if not check the protocols for our desired list of pubsub protocols (gossip, then flood).

vasco-santos: Thanks for the suggestions Jacob!
Mind having a look at the new approach? I think that it encompasses all your thoughts 🙂

src/pubsub.js
```diff
@@ -17,14 +19,24 @@ class BasicPubSub extends Pubsub {
    * @param {String} debugName
    * @param {String} multicodec
    * @param {Object} libp2p libp2p implementation
+   * @param {Object} options
+   * @param {bool} options.fallbackToFloodsub if dial should fallback to floodsub, defaults to true
    * @constructor
    */
-  constructor (debugName, multicodec, libp2p) {
+  constructor (debugName, multicodec, libp2p, options) {
     super(debugName, multicodec, libp2p)
```
jacobheun: Now that we're taking options here, we should pass them to Pubsub so we can pass them down.

```suggestion
     super(debugName, multicodec, libp2p, options)
```

src/pubsub.js
```diff
@@ -50,6 +62,94 @@ class BasicPubSub extends Pubsub {
...
+          multicodec === this.multicodec &&
+          err.code === multistreamConstants.errors.MULTICODEC_NOT_SUPPORTED) {
+          this._dials.add(idB58Str)
+          this.log('dialing %s %s', floodsubMulticodec, idB58Str)
+
+          this.libp2p.dialProtocol(peerInfo, floodsubMulticodec, (err, conn) => {
+            this.log('dial to %s complete', idB58Str)
+
+            // If the dial is not in the set, it means that pubsub has been stopped
+            const pubsubStopped = !this._dials.has(idB58Str)
```
jacobheun: This is being checked in multiple places, what is this checking for? Is it checking if pubsub itself has stopped? If so, we should avoid having `isStopped` logic in here, it should fail gracefully. If it's not, there's likely a better, central solution.

src/pubsub.js
```diff
@@ -50,6 +62,94 @@ class BasicPubSub extends Pubsub {
...
+    }
+
+    // If already dialing this peer, ignore
+    if (this._dials.has(idB58Str)) {
+      this.log('already dialing %s, ignoring dial attempt', idB58Str)
+      return nextTick(() => callback())
+    }
+
+    // Verify if is known that the peer does not support Gossipsub
+    let notSupportingGossipsub = peerInfo.protocols.size && !peerInfo.protocols.has(this.multicodec)
```
jacobheun: I think this might be a little clearer. If it has the floodsub protocol and not gossipsub, we know it doesn't support the latter. I changed the variable name because the negative name was tripping me up reading the code.

```suggestion
    let onlySupportsFloodsub = peerInfo.protocols.has(floodsubMulticodec) && !peerInfo.protocols.has(this.multicodec)
```

src/pubsub.js
```diff
@@ -50,6 +62,94 @@ class BasicPubSub extends Pubsub {
...
+
+    // Define multicodec to use
+    // Should fallback to floodsub if fallback is enabled, protocols were negotiated, and no Gossipsub available
+    let multicodec = this.multicodec
+
+    if (this._options.fallbackToFloodsub && notSupportingGossipsub) {
```
jacobheun: Changing name per my suggestion above

```suggestion
    if (this._options.fallbackToFloodsub && onlySupportsFloodsub) {
```

src/pubsub.js
```diff
@@ -50,6 +62,94 @@ class BasicPubSub extends Pubsub {
...
+  /**
+   * Dial a received peer.
+   * @override
+   * @param {PeerInfo} peerInfo The peer being dialed
+   * @param {function} callback
+   *
+   * @returns {void}
+   */
+  _dialPeer (peerInfo, callback) {
+    const onDial = (idB58Str, pubsubStopped, conn) => {
```
jacobheun: I'm not sure I understand the purpose of this function wrapper for `_onDial`

src/pubsub.js
vasco-santos: I agree with you. This is something based on the base implementation of pubsub: [libp2p/js-libp2p-pubsub/src/index.js#L163-L166](https://github.com/libp2p/js-libp2p-pubsub/blob/master/src/index.js#L163-L166).

I will create an issue on `js-libp2p-pubsub` also mentioning the logic here. I believe that the best approach is to leverage the refactors to async, with the abort dial to use it on stop, which will not need this anymore.

src/pubsub.js
vasco-santos: I moved the logic to the `_onDial` function.

src/pubsub.js
vasco-santos: https://github.com/libp2p/js-libp2p-pubsub/issues/21

jacobheun: So `_dialPeer` is a misnomer (we should change this in the future). We're not actually dialing the peer. This is triggered from the `peer:connect` event, which means we have a muxed connection to the given peer when this is called. We're not dialing to the peer, we're just negotiating the pubsub protocol. 

In the event that pubsub is shut down in the process of that happening, the call to `dialProtocol` will fail, because the switch should also be stopped. Unless users need to stopping pubsub independently of libp2p, this shouldn't be needed.

Unless taking the logic out causes test failures I would just remove it. We can always add a more appropriate `isStarted` check via libp2p/js-libp2p-pubsub#21, and use a `super` call to get the behavior in a future release.

Regardless of that, I don't think this is a blocker for moving forward with this, but it does make the code harder to follow.

Additionally, thinking about this a bit more, I think it would be worth creating a utility method on libp2p to handle the multi dial negotiation for you, especially as the list of protocols grows, so we don't end up with this logic replicated everywhere. Something like:
```js
const protocolPreferences = [gossipsubMulticodec, floodsubMulticodec]
libp2p.newStream(peerInfo, protocolPreferences, callback)
// or
libp2p.dialBest(peerInfo, protocolPreferences, callback)
```

vasco-santos: I like the `dialBest` idea and I will create an issue for that in `js-libp2p`! Other than that, I removed the `pubsubStopped` validation. This should be good to go then

vasco-santos: FYI, created an issue for simplifying this is Gossipsub in a next PR: https://github.com/libp2p/js-libp2p/issues/376

vasco-santos: FYI, created an issue to improve this in a next PR: https://github.com/libp2p/js-libp2p/issues/376

vasco-santos:
@wemeetagain @Mikerah @GregTheGreek can you have a look and merge this PR if you approve it?

GregTheGreek:
> @wemeetagain @Mikerah @GregTheGreek can you have a look and merge this PR if you approve it?

Hey sorry, I'm on it, I know @wemeetagain Is a bit under the weather

src/pubsub.js
```diff
@@ -50,6 +63,76 @@ class BasicPubSub extends Pubsub {
...
+  /**
+   * Dial a received peer.
+   * @override
+   * @param {PeerInfo} peerInfo The peer being dialed
+   * @param {function} callback
+   *
+   * @returns {void}
+   */
+  _dialPeer (peerInfo, callback) {
+    callback = callback || function noop () { }
```
GregTheGreek: is this to prevent executing an undefined function?

src/pubsub.js
```diff
@@ -50,6 +63,76 @@ class BasicPubSub extends Pubsub {
...
+
+            if (err) {
+              this.log.err(err)
+              return callback()
+            }
+            this._onDial(peerInfo, conn, callback)
+          })
+        } else {
+          this.log.err(err)
+          return callback()
```
GregTheGreek: I'm confused about the use of the `callback` in this function. It looks like it's being used as for errors, example:
```
_dialPeer(peerInfo, function() {
    // this is an error
}
```

But typically we wan to pass propagate the error up, and allow the user to handle it. Alternatively we can add two parameters to the the callback `callback(err, value)` and allow the user to handle the error so it looks more like:
```
_dialPeer(peerInfo, function(err) {
    if (err) {
      console.log(err)
      // Do something
    }
}
```
Bare in mind im not familiar with the logger in gossipsub-js but, assuming the log levels are set this could be a silent error.

src/pubsub.js
vasco-santos: exactly!

src/pubsub.js
vasco-santos: You are having good reasoning since we usually use callbacks to propagate values/errors through async logic. However, in this case, is a little different. I will explain to you the flow:

1) Peer got connected (emit event so that everyone interested in it can handle appropriately)
https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L80-L84

2) Pubsub is waiting for peers to connect, in order to add them to the appropriate pubsub router
https://github.com/libp2p/js-libp2p-pubsub/blob/master/src/index.js#L319

3) Gossipsub (in this case) overrides the `_dialPeer` function, which is called and tries to dial it using its pubsub multicodec
(this function here)

All in all, we don't mind if a peer does not join the gossipsub overlay network. We can have peers that are not running any pubsub router and we just log that we could not add them, there is nothing else the peer can do.

src/pubsub.js
```diff
@@ -186,6 +269,10 @@ class BasicPubSub extends Pubsub {
       if (err) {
         return callback(err)
       }
+      // if fallback to floodsub enabled, we need to listen to its protocol
+      if (this._options.fallbackToFloodsub) {
+        this.libp2p.handle('/floodsub/1.0.0', this._onConnection)
```
wemeetagain: consider pulling out the string into a constant.

vasco-santos: wait sorry! I will change it, I already have the constant 😅

vasco-santos:
Should be ready now @wemeetagain 
Let's just wait for CI

src/pubsub.js
GregTheGreek: Hmm ok, thats interesting. So ultimately, we don't care if they can't connect. Makes sense.

vasco-santos:
Thanks ❤️ 
Can I have a new release?

PR #39, opened 2019-07-24
fix: send subscriptions once ready

vasco-santos:
When we send a subscribe or unsubscribe message, we needed to wait for a while in order to guarantee that peers are properly connected. 

This way, we had no precise metrics of when to "believe" that the dial was finished and the peer properly handled by the pubsub protocol. 

In this PR, I added a function to guarantee that when we want to send SUBSCRIBE or UNSUBSCRIBE messages, we wait for peers which are still establishing a stream communication with the peer. I had problems with this when integrating `gossipsub` with `js-ipfs`, as we connect to all the bootstrap peers, and if we start using the pubsub api, we will not be able to write to peers, which are still being connected.

### DIFF
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -367,10 +367,20 @@ class BasicPubSub extends Pubsub {
       this.subscriptions.delete(topic)
     })
 
-    // Broadcast UNSUBSCRIBE to all peers
-    this.peers.forEach((peer) => {
-      peer.sendUnsubscriptions(topics)
-    })
+    // Broadcast UNSUBSCRIBE to all peers ready
+    this.peers.forEach((peer) => sendUnsubscriptionsOnceReady(peer))
+    // make sure that Gossipsub is already mounted
+    function sendUnsubscriptionsOnceReady (peer) {
+      if (peer && peer.isWritable) {
+        return peer.sendUnsubscriptions(topics)
+      }
+      const onConnection = () => {
+        peer.removeListener('connection', onConnection)
+        sendUnsubscriptionsOnceReady(peer)
+      }
+      peer.on('connection', onConnection)
+      peer.once('close', () => peer.removeListener('connection', onConnection))
+    }
 
     this.leave(unTopics)
   }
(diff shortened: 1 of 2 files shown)


src/pubsub.js
```diff
@@ -313,9 +313,19 @@ class BasicPubSub extends Pubsub {
     })
 
     // Broadcast SUBSCRIBE to all peers
-    this.peers.forEach((peer) => {
-      peer.sendSubscriptions(newTopics)
-    })
+    this.peers.forEach((peer) => sendSubscriptionsOnceReady(peer))
+    // make sure that Gossipsub is already mounted
+    function sendSubscriptionsOnceReady (peer) {
```
GregTheGreek: (same comment for `sendUnsubscriptionsOnceReady`)
Do we want this nested inside of `subscribe()` or can it be separated out? Alternatively, does it even need to be a function or can it be a code block within the `forEach`?

vasco-santos: Needs to be a function because it is recursive. `onConnection` we call the function once again.

In a first approach, I extracted and used a single function for both. But this implied that I pass the `peer.sendSubscriptions` / `peer.sendUnsubscriptions` functions via parameters, in order to know which function to pass. The code was not that readable with that approach, and I favored readability instead of less lines of code in this case. It is a total of 5 extra lines of code used, which seemed not a big deal to me.

GregTheGreek: Ahhh I didn't see the recursion - makes sense.

vasco-santos:
@wemeetagain @GregTheGreek 

Rebased the PR after the message validation PR got merged. I think this is good to go

vasco-santos:
@wemeetagain @GregTheGreek can we merge both PRs and do a new release? I am currently on the process of bubbling up gossipsub until `js-ipfs`, but really need this merged and released

GregTheGreek:
I'll get the release out now

vasco-santos:
thank you very much @GregTheGreek

PR #41, opened 2019-07-26
chore: update dependencies

vasco-santos:
Let's have the dependencies updated before the new release


vasco-santos:
@wemeetagain @GregTheGreek can we have a new release? 

Apparently, I also have npm permission, so I can do it if you are ok with that

GregTheGreek:
@vasco-santos thanks for being patient, ive been travelling... i can do this right now!

Issue #43, opened 2019-08-02
Expose option to suppress auto-gossip


wemeetagain:
Currently, all received messages in a subscribed topic are automatically gossiped.
We would like to add an option to _not_ automatically gossip, rather, we would rely on manually re-publishing a message if gossiping is desired.

One known use-case is for validating incoming messages before gossiping.

This probably looks like adding a boolean option, passed in the constructor, similar to how #40 was exposed.

wemeetagain:
Finished w/ #45

PR #45, opened 2019-09-04
Add gossipIncoming option to supress auto-gossip

wemeetagain:
See #43 

Removing auto-gossip allows the handling application to perform validation/transformation/etc to incoming messages before republishing.

One wrinkle with using a republish to gossip is that unchanged messages will be rebroadcast to the original sender (since the `msg.from` and `msg.seqno` has been changed).


test/gossip-incoming.js
```diff
@@ -0,0 +1,140 @@
...
+  let nodeA
+  let nodeB
+  let nodeC
+
+  describe('gossipIncoming == true', () => {
+    beforeEach(async function () {
+      this.timeout(4000)
+      nodeA = await createNode('/ip4/127.0.0.1/tcp/0')
+      nodeB = await createNode('/ip4/127.0.0.1/tcp/0')
+      nodeC = await createNode('/ip4/127.0.0.1/tcp/0')
```
vasco-santos: As a suggestion, I have been starting to use the following for this cases:

```js
[nodeA, nodeB, nodeC] = await Promise.all([
  createNode('/ip4/127.0.0.1/tcp/0'),
  createNode('/ip4/127.0.0.1/tcp/0'),
  createNode('/ip4/127.0.0.1/tcp/0')
])
```

I think it is more readable because we are grouping lines with the same responsibility. Also, this works in parallel which will be faster (as this is a test, it is no more than a suggestion).

I would apply the `Promise.all()` also to the `startNode` below and the `dialNode`

wemeetagain: Thanks for the suggestion :) 
Definitely seems better with the Promise.all.
I'm going to hold off for this PR, as this is somewhat urgent (and we're wanting to cut a new release for our eth2 interop starting tomorrow).
But duely noted!

vasco-santos: go for it 👍

PR #49, opened 2019-10-18
refactor: async migration

vasco-santos:
BREAKING CHANGE: Switch to using async/await and async iterators for all the API. Moreover, gossipsub does not need the libp2p instance anymore, receiving a registerar that enables it to receive the necessary events from libp2p

Needs
- [x] update readme and minor polishments
- [x] [libp2p/js-libp2p-pubsub#26](https://github.com/libp2p/js-libp2p-pubsub/pull/26)
- [x] [libp2p/js-libp2p-floodsub#88](https://github.com/libp2p/js-libp2p-floodsub/pull/88)

With this migration, I also enabled all browser tests except for the gossip ones 🎉


src/pubsub.js
```diff
@@ -245,73 +209,45 @@ class BasicPubSub extends Pubsub {
...
    * @param {Array<rpc.RPC.ControlIHave>} ihave
    * @param {Array<rpc.RPC.ControlIWant>} iwant
    * @param {Array<rpc.RPC.ControlGraft>} graft
    * @param {Array<rpc.RPC.Prune>} prune
-   *
    * @returns {rpc.RPC}
-   *
    */
-  _rpcWithControl (msgs, ihave, iwant, graft, prune) {
+  _rpcWithControl (msgs = [], ihave = [], iwant = [], graft = [], prune = []) {
```
wemeetagain: I think we need to change the calling sites from `rpcWithControl(foo, null, bar, null, null)` to  `rpcWithControl(foo, undefined, bar, undefined, undefined)` to get the same behavior as before.

But the fact that most of the tests pass make me think we may not even need this?

src/pubsub.js
```diff
@@ -1,162 +1,119 @@
...
-const lp = require('pull-length-prefixed')
-const nextTick = require('async/nextTick')
-const { constants: multistreamConstants } = require('multistream-select')
-const { utils } = require('libp2p-pubsub')
-const asyncMap = require('async/map')
+const assert = require('assert')
+const debug = require('debug')
+const debugName = 'libp2p:gossipsub'
+const log = debug(debugName)
+log.error = debug(`${debugName}:error`)
```
wemeetagain: Why aren't we using `this.log` here? Maybe indirectly causing one of the tests to fail.

vasco-santos: So, the issue with using `this.log` is that when we are in the scope of a `for await ...` loop, we can't use `this.x` in that way. I am still looking for a better way of doing this, let's see if I can get a better way

src/index.js
```diff
@@ -12,15 +14,32 @@ const Heartbeat = require('./heartbeat')
...
    */
-  constructor (libp2p, options) {
-    super('libp2p:gossipsub', constants.GossipSubID, libp2p, options)
+  constructor (peerInfo, registrar, options = {}) {
+    assert(PeerInfo.isPeerInfo(peerInfo), 'peer info must be an instance of `peer-info`')
+
+    // registrar handling
+    assert(registrar && typeof registrar.handle === 'function', 'a handle function must be provided in registrar')
+    assert(registrar && typeof registrar.register === 'function', 'a register function must be provided in registrar')
+    assert(registrar && typeof registrar.unregister === 'function', 'a unregister function must be provided in registrar')
```
wemeetagain: Can these three checks be pulled into some sort of utility function, `similarly to PeerInfo.isPeerInfo`?
I suspect there are other users of `registrar`s that could use it.

src/pubsub.js
wemeetagain: I'm not sure I understand. I don't see any logs inside `for await` loops.
My impression is that `this` stays the same even inside that scope. Is there some sort of tricky case where thats not the case?

src/pubsub.js
```diff
@@ -1,162 +1,128 @@
...
-
-    this._dials.add(idB58Str)
-    this.log('dialing %s %s', multicodec, idB58Str)
-
-    this.libp2p.dialProtocol(peerInfo, multicodec, (err, conn) => {
-      this.log('dial to %s complete', idB58Str)
-      this._dials.delete(idB58Str)
-
-      if (err) {
-        // If previously dialed gossipsub and not supported, try floodsub if enabled fallback
```
wemeetagain: Is the floodsub fallback gone in this refactor?

src/index.js
vasco-santos: I will just have this in `libp2p-pubsub` for now. After that, I may add `class-is` to `Registrar`

src/pubsub.js
vasco-santos: I could get this working by using bing

src/pubsub.js
vasco-santos: No! Basically, all the dial logic that existed in this repo, as well as in `floodsub`, `dht` got extracted to an upper layer, as it was mostly replicated.

In this cases, we pass up an array of multicodecs, and it will attempt to dial both `gossipsub` and `floodsub` and call the `onConnect` handler

src/pubsub.js
```diff
@@ -424,15 +370,23 @@ class BasicPubSub extends Pubsub {
...
 
-      callback()
-    })
+  /**
+   * Get the list of topics which the peer is subscribed to.
+   * @override
+   * @returns {Array<String>}
+   */
+  getTopics () {
+    assert(this.started, 'FloodSub is not started')
```
wemeetagain: FloodSub -> PubSub

wemeetagain:
@jacobheun thanks for the heads up.

Sounds like a good plan. :muscle:

vasco-santos:
@wemeetagain @GregTheGreek will you be able to get a release as jacob suggested during the begining of this week?

GregTheGreek:
Yep, I'll get a release under beta Tomorrow

PR #53, opened 2019-12-02
chore: release version v0.2.0

vasco-santos:
@wemeetagain please add the beta tag as you did for the last one


wemeetagain:
published w beta tag

PR #67, opened 2020-03-16
Provide a way to override message id function

tuyennhv:
## Goal
+ Lodestar/eth2 wants a way to specify its own `msgId()` function 
+ The change should be backward compatible

## Details
+ Have default message id function calculated as earlier
+ Provide a way to override that as opttions
+ This is inspired by https://github.com/libp2p/go-libp2p-pubsub/pull/248/files

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,7 @@ Options is an optional object with the following key-value pairs:
 
 * **`fallbackToFloodsub`**: boolean identifying whether the node should fallback to the floodsub protocol, if another connecting peer does not support gossipsub (defaults to **true**).
 * **`emitSelf`**: boolean identifying whether the node should emit to self on publish, in the event of the topic being subscribed (defaults to **false**).
-* **`msgIdFn`**: a function defining the message id given a message, this is optional. For example:
-```js
-const msgIdFn = (message) => Buffer.from(hash(message.data)).toString("base64")
-```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -67,6 +67,10 @@ Options is an optional object with the following key-value pairs:
 
 * **`fallbackToFloodsub`**: boolean identifying whether the node should fallback to the floodsub protocol, if another connecting peer does not support gossipsub (defaults to **true**).
 * **`emitSelf`**: boolean identifying whether the node should emit to self on publish, in the event of the topic being subscribed (defaults to **false**).
+* **`msgIdFn`**: a function defining the message id given a message, this is optional. For example:
```
wemeetagain: Minor wording suggestion:
"a function with signature `(message) => string` defining the message id given a message, used internally to deduplicate gossip (defaults to `(message) => Buffer.from(hash(message.data)).toString("base64")`)"

mpetrunic: It should default to from+seqno

README.md
```diff
@@ -67,6 +67,7 @@ Options is an optional object with the following key-value pairs:
 
 * **`fallbackToFloodsub`**: boolean identifying whether the node should fallback to the floodsub protocol, if another connecting peer does not support gossipsub (defaults to **true**).
 * **`emitSelf`**: boolean identifying whether the node should emit to self on publish, in the event of the topic being subscribed (defaults to **false**).
+* **`msgIdFn`**: a function with signature `(message) => string` defining the message id given a message, used internally to deduplicate gossip (defaults to `(message) => message.from + message.seqno.toString('hex')`)
```
vasco-santos: can you document `messageCache` option?

PR #70, opened 2020-04-07
chore: remove peer-info usage

vasco-santos:
In the context of deprecating `peer-info` as described on [libp2p/js-libp2p#589](https://github.com/libp2p/js-libp2p/issues/589), this PR removes the `peer-info` usage on pubsub internal peer data structure. Moreover, this uses the new topology API, which also uses `peer-id` instead of `peer-info`

BREAKING CHANGE: using new topology api with peer-id instead of peer-info and new pubsub internal peer data structure

Needs:

- [x] [libp2p/js-libp2p-interfaces#42](https://github.com/libp2p/js-libp2p-interfaces/pull/42)
- [x] [libp2p/js-libp2p-pubsub#41](https://github.com/libp2p/js-libp2p-pubsub/pull/41)
- [x] [libp2p/js-libp2p-floodsub#102](https://github.com/libp2p/js-libp2p-floodsub/pull/102)


vasco-santos:
Note: This should be released with a beta tag until we bubble everything up for `js-libp2p` and release everything. Then we will update the tags to latest.

`npx aegir release --type minor --dist-tag beta`

wemeetagain:
Released v0.4.0 w beta tag

PR #74, opened 2020-05-23
Refactor pubsub - remove gossip-specific features

wemeetagain:
Attempting to make the pubsub layer as unopinionated as possible, for possible inclusion in `libp2p-pubsub` eventually.

The pubsub layer now only has minimal defaults to stream data, decode `RPC` objects, process `SubOpt` and `Message` objects, and maintain an updated set of local and peer subscriptions.

This shouldn't effect the pubic gossipsub interface (although a few subtle bugs were noticed/fixed in refactoring).

Major changes:
- rename / refactor rpc handling methods
  - `_processMessages` (perhaps more aptly named `_processRpcStream` or similar) now calls `_decodeRpc` to decode the incoming buffer into an `RPC` object, `_processRpc`, which handles handles a single `RPC` object from a `Peer`
  - `_processRpc` calls `_processRpcSubOpt` for each included subscription and `_processRpcMessage` for each _valid_ included message
- many non-pubsub features moved from pubsub.js
  - creating gossip rpc objects
  - join/leave
  - gossipsub configuration options
  - gossipsub control messages
  - peer shuffling
  - seen message cache

cc @jacobheun


src/shuffle.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
```
vasco-santos, line 1: What do you think you renaming this file to `utils`, or put it inside an `utils` folder?

Some other utils functions might appear and it would be good to have them all together

src/message/index.js
```diff
@@ -7,4 +7,5 @@ const topicDescriptorProto = protons(require('./topic-descriptor.proto.js'))
 
 exports = module.exports
 exports.rpc = rpcProto
```
vasco-santos, line 9: Do we need this export? I agree that the `rcp.RPC` previously used is not good and exporting the RPC message instead of the module is good.

src/message/index.js
```diff
@@ -7,4 +7,5 @@ const topicDescriptorProto = protons(require('./topic-descriptor.proto.js'))
...
+exports.RPC = rpcProto.RPC
 exports.td = topicDescriptorProto
```
vasco-santos, line 11: Should we also just export the message content here?

vasco-santos: I just realised that we do not use this here. This should only be used by the base class `libp2p-pubsub`. Can we also get this out?

src/pubsub.js
```diff
@@ -197,16 +161,42 @@ class BasicPubSub extends Pubsub {
...
 
   /**
-   * @param {rpc.RPC.Message} msg
+   * Handles an subscription change from a peer
+   *
+   * @param {Peer} peer
+   * @param {RPC.SubOpt} subOpt
    */
-  _processRpcMessage (msg) {
+  _processRpcSubOpt (peer, subOpt) {
```
vasco-santos, line 175: We should probably make this name more clear and not abbreviate. What do you think?

src/pubsub.js
```diff
@@ -271,22 +235,31 @@ class BasicPubSub extends Pubsub {
...
     }
 
+    // normalize input and remove existing subscriptions
     topics = utils.ensureArray(topics)
-
     const newTopics = topics.filter((topic) => !this.subscriptions.has(topic))
     if (newTopics.length === 0) {
       return
     }
+    this._subscribe(newTopics)
```
vasco-santos, line 244: Why creating a new function? I think we will not reuse it anywhere else and I don't think it removes complexity.

src/pubsub.js
```diff
@@ -317,22 +284,31 @@ class BasicPubSub extends Pubsub {
...
+  /**
+   * Unsubscribes to topics
+   *
+   * @param {Array<string>} topics
+   * @returns {void}
+   */
+  _unsubscribe (topics) {
     // delete subscriptions
-    unTopics.forEach((topic) => {
+    topics.forEach((topic) => {
```
vasco-santos, line 304: Same as `subscribe`

src/index.js
```diff
@@ -91,6 +107,37 @@ class GossipSub extends BasicPubsub {
     this.heartbeat = new Heartbeat(this)
   }
 
+  /**
+   * Given a topic, returns up to count peers subscribed to that topic
```
vasco-santos, line 111: We can probably get this its own file? What do you think?

src/index.js
```diff
@@ -120,39 +167,51 @@ class GossipSub extends BasicPubsub {
     return peer
   }
 
+  _processRpc (idB58Str, peer, rpc) {
```
vasco-santos, line 170: Please add here jsdocs with the @override annotation

src/index.js
```diff
@@ -120,39 +167,51 @@ class GossipSub extends BasicPubsub {
...
 
-    const outRpc = this._rpcWithControl(iHave, null, iWant, null, prune)
-    this._sendRpc(rpc.from, outRpc)
+    const outRpc = createGossipRpc([], { ihave, iwant, prune })
+    this._sendRpc(peer, outRpc)
   }
 
   /**
    * Process incoming message,
    * emitting locally and forwarding on to relevant floodsub and gossipsub peers
```
vasco-santos, line 201:
```suggestion
   * emitting locally and forwarding on to relevant floodsub and gossipsub peers
   * @override
```

src/index.js
```diff
@@ -327,6 +386,16 @@ class GossipSub extends BasicPubsub {
     this.control = new Map()
   }
 
+  _subscribe (topics) {
```
vasco-santos, line 389: Any reason for not overriding `subscribe` and `unsubscribe`? Also, please add js docs with the @override annotation

src/shuffle.js
wemeetagain: :+1: I just pulled out these utilities w/o much thought to where they should go, will rearrange utils into a `utils` folder

src/message/index.js
wemeetagain: :+1: will remove

src/pubsub.js
wemeetagain: Its reused in src/index.js right now to hook into the "list of new topics" to subscribe to. The idea is that `subscribe` directly receives user input, munges into a canonical form.  `_subscribe` can then be overridden (always calling super._subscribe) and topics can be consumed there without this additional step of filtering/munging of inputs.
We could remove it, but then we will need to do the above filtering again if `subscribe` is overridden. Or perhaps we can have `subscribe` return the list of filtered topics, so overriding functions can still have access.

src/pubsub.js
wemeetagain: Agreed its not a super helpful name. The naming of this, `_processRpc`, `_processRpcMessage`, and `_processRpcControlMessage` were meant to align with the names in the protobuf definition. ie `RPC`, `SubOpt`, `Message`, and `ControlMessage`. Definitely interested in other ideas for consistent naming

src/message/index.js
wemeetagain: :+1:  will remove

src/pubsub.js
vasco-santos: I was thinking about just getting the `subscribe` overridden. I did not see that join needed the topics filtered. I am ok with going like you currently have. I think we should not change the API return values

src/index.js
wemeetagain: :+1:  can move to its own file. In gossipsub v1.1 i think this is going to have to change (return an Array of peers sorted by score instead of a Set)

src/pubsub.js
vasco-santos: I was thinking about expanding the `SubOpt`, but you make a good point. I think that it is a good call to align the namings with the protoBuf definition, rather than make an exception here.

PR #75, opened 2020-05-23
feat: add topic validators to pubsub

wemeetagain:
Clean up validation in pubsub, add topic validator functions
- `validate` method now has additional peer arg and triggers topic-specific validation functions
- topicValidator is `Map<string, (topic, peer, message) => unknown>`
  - topic validators can be added / removed here
- topic validation results can have additional processing overridden with `_processTopicValidatorResult` - default to truthy results passing validation

cc @jacobheun


src/pubsub.js
```diff
@@ -203,6 +209,46 @@ class BasicPubSub extends Pubsub {
...
+   */
+  async validate (peer, message) {
+    const isValid = await super.validate(message)
+    if (!isValid) {
+      return false
+    }
+    for (const topic of message.topicIDs) {
+      const validatorFn = this.topicValidators.get(topic)
+      if (validatorFn) {
+        const result = validatorFn(topic, peer, message)
```
tuyennhv, line 226: it's more likely to me to be a transformer function instead of a validator because the result is not only true/false

wemeetagain: Yeah, esp in gossipsub v1.1, with "extended validators", results can be more than just true/false.
I'm planning on overriding `_processTopicValidatorResult` to handle these "extended validators". Other protocols might have different handling, so they can ofc override with protocol-specific handling.

The reason I named it 'validator' rather than transformer is to align with the gossipsub v1.0 and v1.1 spec. (see here: https://github.com/libp2p/specs/tree/master/pubsub#topic-validation and here: https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#extended-validators)

tuyennhv: thanks, that's good to me

wemeetagain:
> Is there any reason for not adding this to libp2p-pubsub?

I've been viewing these two open PRs (#75 and #74) as sort of clean up tasks so that this `src/pubsub.js` may be merged into `libp2p-pubsub`. (And so that this pubsub impl "has what we need" for the gossipsub v1.1 upgrade)
I think for this one, one of the bigger reasons I didnt' PR directly into libp2p-pubsub is bc this would be a breaking API change to the `validate` function. (from `async (message: RPC.Message) => boolean` to `async (peer: Peer, message: RPC.Message) => boolean`)

Somehow it seemed simpler to explore this with my pseudo libp2p-pubsub impl before having the broader discussion about merging breaking functionality back upstream.

Happy to make the PR to libp2p-pubsub if you think thats better tho.

vasco-santos:
@wemeetagain seems a good approach. We can start here and then iterate to `js-libp2p-pubsub` then! I will make sure to review this PR tomorrow

src/pubsub.js
```diff
@@ -203,6 +209,46 @@ class BasicPubSub extends Pubsub {
     this._handleRpcControl(peer, rpc)
   }
 
+  /**
+   * Validates the given message.
+   * @param {Peer} peer
+   * @param {rpc.RPC.Message} message
+   * @returns {Promise<Boolean>}
+   */
+  async validate (peer, message) {
```
vasco-santos, line 218:
```suggestion
  async validate (message, peer) {
```

In a semantic perspective, I think that the message should come first as it is what is being validated. Moreover, if we make the message the first parameter and validate if we receive a peer, this will not be a breaking change. We will probably want to add this feature to `floodsub` afterwards, but I think that we should not make this a breaking change, since the topicValidators are also not required.

src/pubsub.js
```diff
@@ -203,6 +209,46 @@ class BasicPubSub extends Pubsub {
...
+    const isValid = await super.validate(message)
+    if (!isValid) {
+      return false
+    }
+    for (const topic of message.topicIDs) {
+      const validatorFn = this.topicValidators.get(topic)
+      if (validatorFn) {
+        const result = validatorFn(topic, peer, message)
+        if (!this._processTopicValidatorResult(topic, peer, message, result)) {
```
vasco-santos, line 227: return the function result instead of the condition?

src/pubsub.js
```diff
@@ -203,6 +209,46 @@ class BasicPubSub extends Pubsub {
...
+   *
+   * Defaults to true if truthy
+   *
+   * Override this method to provide custom topic validator result processing (eg: scoring)
+   *
+   * @param {String} topic
+   * @param {Peer} peer
+   * @param {rpc.RPC.Message} message
+   * @param {unknown} result
+   * @returns {Promise<Boolean>}
```
vasco-santos, line 246:
```suggestion
   * @returns {boolean}
```

src/pubsub.js
```diff
@@ -78,6 +78,15 @@ class BasicPubSub extends Pubsub {
      * @returns {string} message id as string
      */
     this.defaultMsgIdFn = (msg) => utils.msgId(msg.from, msg.seqno)
+
+    /**
+     * Topic validator map
+     *
+     * Keyed by topic
+     * Topic validators are functions with the following input:
+     * (topic: string, peer: Peer, message: rpc.RPC)
```
vasco-santos, line 87:
```suggestion
     * @type {Map<string, validator>}
```

You can define a typedef for the validator as below:

```
/**
 * Topic validator function.
 * @typedef {function(topic: string, peer: Peer, message: rpc.RPC): boolean} validator
 */
```

src/pubsub.js
wemeetagain: Do you see the topicValidators getting added to `libp2p-pubsub`? I was under the impression that our current `libp2p-pubsub` implementation is not _quite_ up to spec.
https://github.com/libp2p/specs/tree/master/pubsub#topic-validation
Makes it seem like topic-specific validators should be there. I guess its a little ambiguous, you could say that overriding the existing `validate` method is enough to fulfill the spec requirements.

vasco-santos: Yes, I totally see the topicValidators getting added to `libp2p-pubsub`, but pubsub routers should not be required to used them. At least for now, once we update this in floodsub, we can decide to make it required. That's why I recommended switching the parameters and validate if we have a peer, this way this could be added to libp2p-pubsub but it would not required floodsub to be updated

src/pubsub.js
wemeetagain: process the results all at once w a reduce instead of bailing after the first falsy processed result?

vasco-santos: I think that this was changed according to what I was thinking :)

PR #77, opened 2020-05-29
Refactor to use Typescript

wemeetagain:
Refactored to use typescript with minimal changes to process and code.
Imo, using typescript in this project will pay dividends as the gossipsub feature-set grows. It becomes much easier to reason about this codebase which has many moving parts and datatypes when there's that minimal level of type/sanity checking.
That said, I attempted to do this update in a way that's minimally invasive to the standard libp2p workflow.

- typescript code lives in `ts/`, gets transpiled to `src/`
  - `prebuild` and `pretest` scripts runs `tsc`
  - build step left untouched - `aegir build`
  - emits js code in strict mode
  - pubsub.js left untouched (expected to migrate to js-libp2p-pubsub repo eventually)
  - `src/` not checked into git, like `dist/`
- built `dist/index.min.js` is 2KB larger, from 468KB to 470KB
- linting now performed by `eslint` (using `standard` and `@typescript-eslint` configs/plugins) instead of `aegir`, as the typescript files themselves are to be linted
- tests left untouched
- caught a few bugs
  - doc strings out of sync
  - no null checks for nullish values
  - improper use of functions due to misleading variable names
- `GossipSub` => `Gossipsub` throughout (I can remove to future PR if desired)
- types in `ts/peer.ts` should eventually go to `js-libp2p-pubsub`


.eslintrc.js
```diff
@@ -0,0 +1,34 @@
...
+        'delimiter': 'none',
+        'requireLast': false
+      },
+      'singleline': {
+        'delimiter': 'comma',
+        'requireLast': false
+      }
+    }],
+    '@typescript-eslint/indent': 'off', // This is the job of StandardJS, they are competing rules so we turn off the Typescript one.
+    'no-unused-vars': 'off', // On the fence about using this one, sometimes we import a package that is never used directly.
```
mpetrunic, line 29: I think you can enable that and use `import "<package>";` for those kind of imports

ts/getGossipPeers.ts
```diff
@@ -12,16 +12,16 @@ const { shuffle } = require('./utils')
  * @returns {Set<Peer>}
  *
  */
-module.exports = function getGossipPeers (router, topic, count) {
+export function getGossipPeers (router: Gossipsub, topic: string, count: number): Set<Peer> {
```
mpetrunic, line 15: This is not equal to original.
`module.exports =  function getGossipPeers` should be exported as  `export = function getGossipPeers` to not introduce breaking change if somebody is using this directly

ts/message/index.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export interface Message {
+  from?: Buffer
+  data?: Buffer
+  seqno?: Buffer
+  topicIDs: string[]
+  signature?: Buffer
+  key?: Buffer
+}
```
mpetrunic, line 18: Are all those really optional? you can use "Partial<Message>" if you need to initialize partially

ts/message/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Same as Message, but `from` is an optional string
+ */
+export interface InMessage {
```
mpetrunic, line 23: https://stackoverflow.com/questions/43159887/make-a-single-property-optional-in-typescript to avoid duplication

ts/index.ts
```diff
@@ -674,10 +704,10 @@ class GossipSub extends BasicPubsub {
...
    */
-  _now () {
+  _now (): number {
     return Date.now()
   }
 }
 
-module.exports = GossipSub
-module.exports.multicodec = constants.GossipSubID
+module.exports = Gossipsub
```
mpetrunic, line 712: export = Gossipsub should be used in typescript
https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

ts/peer.ts
```diff
@@ -0,0 +1,29 @@
...
+  sendMessages (msgs: Message[]): void
+  updateSubscriptions (subOpts: SubOpts[]): void
+  close (): void
+}
+
+export interface Registrar {
+  handle (): void
+  register (): void
+  unregister (): void
+}
```
mpetrunic, line 29: You should probably update those interfaces in libp2p-ts and use from there

ts/message/index.ts
wemeetagain: Yeah, they're optional. Its following the protobuf definition, we should call that out explicitly in the comments

.eslintrc.js
wemeetagain: This line was just copied from the link at the top of that file, will delete

ts/getGossipPeers.ts
wemeetagain: I'm going to leave this one as is. Its not part of the exported interface, so there's not any expectation of backwards compatibility here.

ts/peer.ts
wemeetagain: I want to upstream these to js-libp2p-pubsub directly :)
Leaving here for now, but will open a PR to add ts declarations there soon.

package.json
```diff
@@ -10,8 +10,9 @@
   ],
   "types": "src/index.d.ts",
   "scripts": {
-    "lint": "aegir lint",
+    "lint": "eslint --ext .ts ts",
     "release": "aegir release",
+    "prebuild": "tsc",
     "build": "aegir build",
     "test": "aegir test",
```
vasco-santos, line 17: shoul we add the prebuild step to the test script? Otherwise, it will fail until we run it.

> open question - should src/ be checked into git? right now it is not

This would solve this issue. But this is not expected in a typescript project, right?

wemeetagain: Yeah, good idea, will add a `pretest` script. In the PR currently, I've manually run `prebuild` before each test step in travis, not a very good solution.

In adding `src/` to version control, I'd just be worried about keeping it in sync. I think that's one reason its not checked in, similar to how we don't check in the `dist/` folder

wemeetagain:
@vasco-santos Thanks for bringing up the larger issue of standardizing the tooling and strategy for supporting typescript across the js-libp2p/js-ipfs ecosystem.

I do think that using full typescript here will be beneficial, over, eg: autogen declarations. Imo it mainly boils down to improved dev tooling and type-checking sanity checks. I think the benefits outweigh the additional machinery and differences with other libp2p repos.

The current gossipsub codebase has many rough edges and using typescript here clears up many ambiguities. As part of this process, several bugs were found just from having type-checking. While these bugs were obvious in hindsight, those particular issues would have never passed the base level of scrutiny we get using typescript. I'd love to see this gossipsub implementation be polished and be easily shown to be correct, and I think type-checking plays a role here.

The dev tooling aspect is also useful to call out. I've been working on gossipsub v1.1 in a local branch, and having the full IDE type-checking/hinting and autocompletion significantly reduces the mental burden of eg: spelling out logic with app-specific datastructures. We're able to declare datastructures with interfaces (in the case of gossipsubv1.1, these interfaces can be fairly large and wordy) and seamlessly use these to aid development without having to navigate opaque objects by hand. Using vanilla js, I was struggling to have my IDE link JSDocs throughout the project, and having to manually check for typos and correct usage, so the experience was not nearly so pleasant. 

Remaining thoughts are how to best do this PR w/o negatively impacting the js-libp2p ecosystem and with maximal standardization/collaboration with the js-libp2p ecosystem.

I would love to keep using aegir where possible, (using it for linting eventually, if/when supported), and the current structure is an attempt to simply extend the current processes to a typescript codebase. Ideally after ts => js transpilation/pre-processing, "things can be treated the same".

Perhaps this may be the only js-libp2p repo using typescript, but I would like to get this right to maximally align with the whole ecosystem so it doesn't have to be treated strangely and can benefit from standardization.

vasco-santos:
@wemeetagain thanks for exposing your thoughts. I agree with your thoughts and we should move on the way you are working on!

vasco-santos:
FYI, just released `0.4.3` with this PR and a custom aegir on my local machine to skip the dirty validation

PR #78, opened 2020-06-03
fix: use unidirectional streams

vasco-santos:
This PR makes libp2p pubsub subsystem use unidirectional streams instead of bidirectional streams per discussion on [ipfs/go-ipfs#7390](https://github.com/ipfs/go-ipfs/issues/7390).

More details about the reasoning for this can be found in https://discuss.libp2p.io/t/gossip-questions/257/6

I have tested this change on all libp2p repos: `js-libp2p`, `libp2p-daemon` and `interop` and everything is working properly. The tests here need to be updated though, as a result of the tests being implemented considering a bidirectional stream.

Bear in mind that a lot of this "connection logic" can be abstracted. I did not worry about it and just followed what we already had in floodsub and gossipsub. Hopefully, after `gossipsub@1.1`, I want to create the libp2p pubsub interface and move/refactor the tests there and re-use most of them in all our router implementations.

Needs:

- [ ] https://github.com/libp2p/js-libp2p-pubsub/pull/45

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "err-code": "^2.0.0",
     "it-length-prefixed": "^3.0.0",
     "it-pipe": "^1.0.1",
-    "libp2p-pubsub": "^0.5.0",
+    "libp2p-pubsub": "libp2p/js-libp2p-pubsub#fix/use-unidirectional-streams",
     "p-map": "^4.0.0",
     "peer-id": "~0.13.3",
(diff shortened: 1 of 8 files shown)


vasco-santos:
Per https://github.com/ipfs/go-ipfs/issues/7390#issuecomment-638298041 this should be ready to go

cc @achingbrain cannot ask you direct review

PR #87, opened 2020-06-16
gs1.1 Integrate peer score

wemeetagain:
Building off #83 
Minimal integration of peer score
- adds a PeerScore object to gossipsub
- plugs into gossipsub `start`/`stop`/`_addPeer`/`_removePeer`
- pass a ConnectionManager into the constructor of gossipsub
- update tests to new constructor

### DIFF
--- a/ts/score/peerScore.ts
+++ b/ts/score/peerScore.ts
@@ -16,7 +16,7 @@ interface Connection {
   remotePeer: PeerId
 }
 
-interface ConnectionManager {
+export interface ConnectionManager {
   getAll(id: string): Connection[]
   // eslint-disable-next-line @typescript-eslint/ban-types
   on(evt: string, fn: Function): void
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "err-code": "^2.0.0",
     "it-length-prefixed": "^3.0.0",
     "it-pipe": "^1.0.1",
-    "libp2p-pubsub": "~0.5.2",
+    "libp2p-pubsub": "~0.5.3",
     "p-map": "^4.0.0",
     "peer-id": "~0.13.12",
     "protons": "^1.0.1",
--- a/test/2-nodes.spec.js
+++ b/test/2-nodes.spec.js
@@ -14,6 +14,7 @@ const {
   createGossipsubNodes,
   createGossipsubConnectedNodes,
   mockRegistrar,
+  mockConnectionManager,
   expectSet,
   ConnectionPair,
   first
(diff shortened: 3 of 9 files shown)


ts/score/peerScore.ts
```diff
@@ -77,7 +83,7 @@ export class PeerScore {
     }
     clearInterval(this._backgroundInterval)
     delete this._backgroundInterval
-    this._addressBook.off('change:multiaddrs', this._updateIPs)
+    this._connectionManager.off('change:multiaddrs', this._updateIPs)
```
vasco-santos, line 86: Should we just remove the IPs instead?
If I understood correctly, this will keep them in the data structure

wemeetagain: ah, yeah, we should probably clear all IPs.
This is also wrong, bc the ConnectionManager emits the `'peer:connect'` and `'peer:disconnect'` events, not `'change:multiaddrs'` events

wemeetagain: it seems that there are some problems with our current approach of updating IPs, and it seems the events currently emitted by the connection manager are not sufficient for tracking IPs to open connections for peers with scores.

issue 1: `ConnectionManager` doesn't emit on closed connections. `ConnectionManager#_maybeDisconnectOne`, doesn't emit when a connection is closed, more generally, `ConnectionManager` isn't tracking/emitting on manually closed connections, but rather only on calls to `ConnectionManager#onDisconnect`.

issue 2: `ConnectionManager` emits `peer:connect` event before storing the emitted connection.
The current approach is to use `ConnectionManager#getAll` on each `peer:connect` event for a list of current IPs, but since the event is triggered before the new connection is stored, the call to `getAll` will not include the new connection unless the event handler awaits the `ConnectionManager#onConnect` finishing in its entirety. But that potentially also runs into issues re issue 1.

I think, for now, to move forward, we should fall back to calling `ConnectionManager#getAll` for each peer in the `background` method here, in the same synchronous fashion that go-libp2p-pubsub does, while we sort out what a good solution for asynchronously tracking open connection IPs would look like.

wemeetagain: whoops, i thought this was #83, i've updated that PR with the idea ^
Once we merge, the diff here should be tiny

vasco-santos: I had a look and I think we can go with this for now! I will create an issue for tracking this connectionManager issues that you mentioned. Hopefully we can get to that in `libp2p@0.30`. Meanwhile, we should probably create an issue in gossipsub to sign that this should be changed once libp2p connection manager properly supports these needs.

PR #89, opened 2020-06-17
gs1.1 flood publishing

wemeetagain:
Builds off #87 
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#flood-publishing

Add `floodPublish` option, defaulted to true
With flood publishing on:
Self-published messages should be sent to all peers with scores above the publish threshold

With flood publishing off:
Self-published messages should be sent using gossipsub 1.0 functionality

### DIFF
--- a/ts/getGossipPeers.ts
+++ b/ts/getGossipPeers.ts
@@ -1,27 +1,38 @@
-import * as constants from './constants'
+import { GossipsubIDv10, GossipsubIDv11 } from './constants'
 import { shuffle } from './utils'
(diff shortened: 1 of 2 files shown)


wemeetagain:
I'm going to update `getGossipPeers` in a separate PR first for use here.
Looking at the go implementation, the gossipsub v1.0 behavior is also tweaked, uses the publish threshold to select mesh peers. This (and other) features will benefit `getGossipPeers` to be updated to suit

PR #91, opened 2020-06-18
gs1.1 outbound mesh quota

wemeetagain:
Implements
https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#outbound-mesh-quotas
https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#heartbeat-maintenance
https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#spam-protection-measures `GRAFT` spam protection

Most of these edits made directly in line with the go-libp2p-pubsub impl. https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go
I tried to avoid rewriting / adding all features, so there are some conditionals/blocks that aren't yet included (eg: direct peers, graft/prune backoff tracking, px, etc)

Big updates:
`_addPeer`, `_removePeer`
- add/remove `outbound` cache for peer

`_handleGraft`
- update to ignore negative score peer grafts
- only accept outbound grafts if over Dhi

`heartbeat`
- cache score
- graftPeer, prunePeer inline functions
- heartbeat maintenance and outbound mesh quota additions

`test/gossip.js`
- tweak second test to be... better

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -42,7 +42,9 @@ class Gossipsub extends BasicPubsub {
   lastpub: Map<string, number>
   gossip: Map<Peer, ControlIHave[]>
   control: Map<Peer, ControlMessage>
+  outbound: Map<Peer, boolean>
   score: PeerScore
+  _connectionManager: ConnectionManager
   _options: GossipOptions
 
   public static multicodec: string = constants.GossipsubIDv10
@@ -137,6 +139,13 @@ class Gossipsub extends BasicPubsub {
     this.control = new Map()
 
     /**
+     * Connection direction cache, marks peers with outbound connections
+     *
+     * @type {Map<Peer, boolean>}
+     */
+    this.outbound = new Map()
+
+    /**
      * Use the overriden mesgIdFn or the default one.
      */
     this._msgIdFn = options.msgIdFn || this.defaultMsgIdFn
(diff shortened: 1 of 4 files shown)


ts/index.ts
```diff
@@ -170,6 +184,18 @@ class Gossipsub extends BasicPubsub {
     // Add to peer scoring
     this.score.addPeer(peerId.toB58String())
 
+    // track the connection direction
+    let outbound = false
+    for (const c of this._connectionManager.getAll(peerId)) {
+      if (c.stat.direction === 'outbound') {
+        if (Array.from(c.registry.values()).some(rvalue => protocols.includes(rvalue.protocol))) {
```
wemeetagain, line 191: This isn't great, not sure what the best way to do this is.
Trying to ensure that gossipsub is one of the protocols of one of the streams on the connection (considering that there may be multiple connections for a peer, not all containing a gossipsub stream)

ts/score/peerScore.ts
```diff
@@ -14,10 +14,14 @@ const log = debug('libp2p:gossipsub:score')
...
   remotePeer: PeerId
+  stat: {
+    direction: 'inbound' | 'outbound'
+  }
+  registry: Map<string, {protocol: string}>
 }
 
 export interface ConnectionManager {
-  getAll(id: string): Connection[]
+  getAll(peerId: PeerId): Connection[]
```
wemeetagain, line 24: this was a mistake that got through, fixed here and below

ts/index.ts
vasco-santos: humm, yeah!
Maybe the connection could provide an utility API to check if it has open streams using a given protocol.
I will add to my TODO list to consider and implement it. Meanwhile, we can go this way

Issue #92, opened 2020-06-19
_processRpcMessage lifecycle


wemeetagain:
We recently moved the `seenCache` to the gossipsub layer, this is a sanity check to see if this is good.
In reading thru the go-libp2p-pubsub code, I see that their `seenCache` is actually at the pubsub layer!
Their processing goes in the following order ([here](https://github.com/libp2p/go-libp2p-pubsub/blob/8945f91465bc3ba418688684a2c29c384cd79647/pubsub.go#L924)):
1. filter blacklisted peers
2. filter unsigned messages in strictSigning mode
3. filter messages sent by others claiming to be from self
4. filter seenCache messages
5. push message onto topic validation queue, filter if queue full
6. add message to seen cache
7. process success case

We don't need to mimic the go-libp2p-pubsub implementation _exactly_ but we should probably respect the behavior as closely as possible.

PR #93, opened 2020-06-19
gs1.1 extended validators

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#extended-validators

Add extended topic validators and minimal refactoring to properly support it
Integrate scoring for validation (validation begin/reject/ignore), and successful message delivery

`src/pubsub.js` required some refactoring to support this feature. The main reason is that messages should first be checked for `seenCache` entries before running topic validation. Since the `seenCache` is currently on the gossipsub layer, and validation on the pubsub layer, there was some reworking to allow `seenCache` to checked before validation -- namely, moving validation from `_processRpc`, as a pre-step to `_processRpcMessage`, to instead run validation within `_processRpcMessage`. (See https://github.com/libp2p/js-libp2p-pubsub/issues/50 and #92 for wrinkles still lingering) 

This required some refactoring of the pubsub tests within this repo.

`src/pubsub.js` also got some slight refactoring around emitting messages.
`_emitMessage` now takes a single parameter (a InMessage), and is reused in both self publishing and publishing from received messages.
`_processRpcMessage` was extended with an additional method `_publishFrom`, which is meant to act as the "success case" for messages received from peers (step 7 in #92). This method is now extended on the gossipsub layer to broadcast messages to floodsub/mesh peers and support "message delivered" peer scoring.

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -8,6 +8,7 @@ import {
   ControlMessage, ControlIHave, ControlGraft, ControlIWant, ControlPrune
 } from './message'
 import * as constants from './constants'
+import { ExtendedValidatorResult } from './constants'
 import { Heartbeat } from './heartbeat'
 import { getGossipPeers } from './getGossipPeers'
 import { createGossipRpc, shuffle, hasGossipProtocol } from './utils'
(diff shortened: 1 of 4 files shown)


ts/index.ts
```diff
@@ -280,49 +281,92 @@ class Gossipsub extends BasicPubsub {
...
-      if (peer.protocols.includes(constants.FloodsubID) &&
-        peer.id.toB58String() !== msg.from &&
-        utils.anyMatch(peer.topics, topics) &&
-        peer.isWritable
-      ) {
-        peer.sendMessages(utils.normalizeOutRpcMessages([msg]))
-        this.log('publish msg on topics - floodsub', topics, peer.id.toB58String())
-      }
-    })
+    if (this._options.gossipIncoming) {
```
vasco-santos, line 310: I would rather short circuit here than have all the remaining lines with more indentation. What do you think?

ts/pubsub.js
```diff
@@ -248,17 +235,43 @@ class BasicPubSub extends Pubsub {
...
+    this._publishFrom(peer, msg)
+  }
 
+  /**
+   * Publish a message sent from a peer
+   * @param {Peer} peer
+   * @param {InMessage} msg
+   * @returns {void}
+   */
+  _publishFrom (peer, msg) {
```
vasco-santos, line 263: Should we validate if we should emit to self here?

ts/index.ts
wemeetagain: The reason I changed this is to still allow for `super._publishFrom` to be called at the bottom of the function, regardless of `gossipIncoming`, L340 below.
I forgot to put this in the description above, but its a slight change in behavior.
The reason for calling `super._publishFrom` is to have this ordering, which was switched before:
1. (optionally) publish incoming message to floodsub/gossipsub peers (`this._publishFrom`)
2. (optionally) emit incoming message as an event (`super._publishFrom`)

Reasoning for this is that iirc, event handlers are executed synchronously as events are emitted, which seems like it might interfere with the timing of publishing incoming messages. I guess the general rule I've been using is to try to have events emitted at the end of processing.

ts/pubsub.js
```diff
@@ -393,23 +406,22 @@ class BasicPubSub extends Pubsub {
...
-    const msgObjects = await pMap(messages, buildMessage)
+    const msgObjects = messages.map(buildMessage)
 
+    // Emit to self if I'm interested and emitSelf enabled
+    this._options.emitSelf && msgObjects.forEach(msg => this._emitMessage(msg))
+
+    const signMessage = (msg, cb) => this._buildMessage(msg)
     // send to all the other peers
-    this._publish(utils.normalizeOutRpcMessages(msgObjects))
+    this._publish(await pMap(msgObjects, signMessage))
```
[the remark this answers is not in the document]
wemeetagain: This was another thing I forgot to mention in the description.
Changed the function signature to accept an `InMessage`. (gossipsub._publish was already accepting `InMessage`, and normalizing there, this was an undiagnosed type error)

Changed in this direction (rather than changing `gossipsub._publish`), related to https://github.com/libp2p/js-libp2p-pubsub/issues/50, I think it makes sense to operate on `InMessage` when working with "message" objects whenever possible.

ts/pubsub.js
wemeetagain: We weren't doing this before, I believe the current behavior of `emitSelf` is to suppress the events from messages published from self. The naming here is confusing :(

ts/pubsub.js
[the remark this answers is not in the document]
vasco-santos: sounds good!

ts/pubsub.js
vasco-santos: you are right. We can also discuss the naming later if you have any suggestion

ts/index.ts
vasco-santos: it makes sense!

PR #95, opened 2020-06-23
gs1.1 pruning backoff

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange prune backoff, not peer exchange

Follows the go-libp2p-pubsub implementation closely.

Extends `ControlPrune` protobuf message and interface to include "peer info"s and backoff. (For now, only backoff is used)
Adds `backoff` Map, a map of maps, indexed by topic, then peer, with a value of a timestamp. If a peer sends a graft for a topic while a backoff entry exists, immediately prune, re-extend the backoff period.
Adds a `heartbeatTicks` counter, will be used for opportunistic grafting in a future PR. Here it's used to clear old backoff entries once every 15 heartbeats.
Adds a `_makePrune` `ControlPrune` constructor. For now, the PX section is not there.

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -42,8 +42,10 @@ class Gossipsub extends BasicPubsub {
   lastpub: Map<string, number>
   gossip: Map<Peer, ControlIHave[]>
   control: Map<Peer, ControlMessage>
+  backoff: Map<string, Map<string, number>>
   outbound: Map<Peer, boolean>
   score: PeerScore
+  heartbeatTicks: number
   _connectionManager: ConnectionManager
   _options: GossipOptions
 
@@ -139,6 +141,11 @@ class Gossipsub extends BasicPubsub {
     this.control = new Map()
 
     /**
+     * Prune backoff map
+     */
+    this.backoff = new Map()
+
+    /**
      * Connection direction cache, marks peers with outbound connections
      *
      * @type {Map<Peer, boolean>}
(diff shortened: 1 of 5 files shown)


ts/index.ts
```diff
@@ -467,6 +497,65 @@ class Gossipsub extends BasicPubsub {
...
+    }
+  }
+
+  /**
+   * Clear expired backoff expiries
+   * @returns {void}
+   */
+  _clearBackoff (): void {
+    // we only clear once every 15 ticks to avoid iterating over the maps too much
+    if (this.heartbeatTicks % 15 !== 0) {
```
vasco-santos, line 546: Where does the 15 come from? Should this be a constant/configuration option?

vasco-santos: Also, should we reset it instead?

wemeetagain: 15 is just pulled from go-libp2p-pubsub. I'll pull it into a constant.

We can't reset it, at least not here, because other features will be relying on the ticks in a similar way (eg: https://github.com/ChainSafe/js-libp2p-gossipsub/pull/96/files#diff-86c7c8f2b553297cee358fb131db4f7bR239)
In go-libp2p-pubsub, this counter is never reset, even on `stop`. Shouldn't be a problem for us though, even at a heartbeat every 100ms, we should have over 10 million years before getting out of safe integer range.

PR #96, opened 2020-06-23
gs1.1 opportunistic grafting

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#opportunistic-grafting

Depends on #95 
Follows go-libp2p-pubsub implementation

Only one block of code in heartbeat added

### DIFF
--- a/ts/heartbeat.ts
+++ b/ts/heartbeat.ts
@@ -235,6 +235,39 @@ export class Heartbeat {
         }
       }
 
+      // should we try to improve the mesh with opportunistic grafting?
+      if (this.gossipsub.heartbeatTicks % constants.GossipsubOpportunisticGraftTicks === 0 && peers.size > 1) {
+        // Opportunistic grafting works as follows: we check the median score of peers in the
+        // mesh; if this score is below the opportunisticGraftThreshold, we select a few peers at
+        // random with score over the median.
+        // The intention is to (slowly) improve an underperforming mesh by introducing good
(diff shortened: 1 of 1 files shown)


ts/heartbeat.ts
```diff
@@ -217,13 +226,48 @@ export class Heartbeat {
...
-            // filter our current mesh peers and peers with negative score
-            return !peers.has(p) && getScore(p.id.toB58String()) >= 0
+            const id = p.id.toB58String()
+            // filter our current mesh peers, peers we are backing off, peers with negative score
+            return !peers.has(p) && (!backoff || !backoff.has(id)) && getScore(id) >= 0
           }).forEach(graftPeer)
         }
       }
 
+      // should we try to improve the mesh with opportunistic grafting?
```
wemeetagain, line 238: This is the block

ts/heartbeat.ts
```diff
@@ -235,6 +235,39 @@ export class Heartbeat {
...
```
vasco-santos, line 238:
```suggestion
```
This can go away, right?

wemeetagain: This is just a commentary about the block of code, pulled from https://github.com/libp2p/go-libp2p-pubsub/blob/8945f91465bc3ba418688684a2c29c384cd79647/gossipsub.go#L1387
I suggest we keep it

PR #99, opened 2020-06-30
gs1.1 direct connect

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#explicit-peering-agreements

Follows the go-libp2p-pubsub implementation closely

### DIFF
--- a/ts/heartbeat.ts
+++ b/ts/heartbeat.ts
@@ -90,6 +90,9 @@ export class Heartbeat {
     // clean up expired backoffs
     this.gossipsub._clearBackoff()
 
+    // ensure direct peers are connected
+    this.gossipsub._directConnect()
+
     // maintain the mesh for topics we have joined
     this.gossipsub.mesh.forEach((peers, topic) => {
       // prune/graft helper functions (defined per topic)
@@ -151,8 +154,8 @@ export class Heartbeat {
         const ineed = constants.GossipsubD - peers.size
         const peersSet = getGossipPeers(this.gossipsub, topic, ineed, p => {
           const id = p.id.toB58String()
-          // filter out mesh peers, peers we are backing off, peers with negative score
-          return !peers.has(p) && (!backoff || !backoff.has(id)) && getScore(id) >= 0
+          // filter out mesh peers, direct peers, peers we are backing off, peers with negative score
+          return !peers.has(p) && !this.gossipsub.direct.has(id) && (!backoff || !backoff.has(id)) && getScore(id) >= 0
         })
 
         peersSet.forEach(graftPeer)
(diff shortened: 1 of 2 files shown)


vasco-santos:
> All js-libp2p interfaces collected into ts/interfaces.ts (This should likely be refactored away into a separate library eventually, eg: into js-libp2p directly or chainsafe/libp2p-ts. But for now, its extremely convenient to tweak the interfaces as needed all in one place.)

I am ok on having this for now. But I would like to have us to generate the types declarations via js docs with aegir

> gossipsub constructor signature changed significantly, now (libp2p: Libp2p, options: object)
This will require a significant rewrite of all existing tests, at least the sections that create new gossipsub objects. I've begun to do this, adding libp2p, libp2p-websockets, libp2p-noise, libp2p-mplex as dev dependencies. Unfortunately, I've run into issues with browser compatability, so just decided to push what I have for now. Hopefully we can discuss and figure out how to proceed.

You will need to have a signal server for the tests. Check it out this example:
- https://github.com/libp2p/js-libp2p/blob/master/.aegir.js - this runs a node peer that browsers will interact with

This example allows us to test browser considering that a browser node will connect to it directly. I would recommend that we setup here this kind of server but with the server being a circuit relay. We would spin up n nodes, that would listen on the circuit relay peer for new connections from the other peers. Then, we can have configuration files for browser and node like here: https://github.com/libp2p/js-libp2p/tree/master/test/utils , but I would also have a function there to compute the listen multiaddrs to use. That is, we do not need to connect via the relay in node, just when running browser tests.
What do you think?

ts/pubsub.js
```diff
@@ -40,8 +36,8 @@ class BasicPubSub extends Pubsub {
     super({
       debugName,
       multicodecs,
-      peerId,
-      registrar,
+      peerId: libp2p.peerId,
+      registrar: libp2p.registrar,
```
vasco-santos, line 40: When we refactor pubsub, I think we should also send libp2p to it. What do you think?

wemeetagain: for sure :+1:

wemeetagain:
> Maybe a new PR just adding libp2p as a parameter to have the tests good before getting this in?

Yeah, I'll split this up into two PRs:

first collecting interfaces and adding libp2p as a parameter, including passing tests,
second adding direct peer connections

vasco-santos:
@wemeetagain thanks 👍 

For the tests, you can get ideas from my WIP work for the rendezvous:
server: https://github.com/libp2p/js-libp2p-rendezvous/blob/feat/rendezvous-protocol-full-implementation/.aegir.js
https://github.com/libp2p/js-libp2p-rendezvous/blob/feat/rendezvous-protocol-full-implementation/test/utils.js#L38

It is working over circuit-relay for connecting the nodes

wemeetagain:
Ok, awesome, I'll give it all a shot. Thanks for the examples, super helpful.

PR #100, opened 2020-07-01
gs1.1 peer exchange

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange

Follows go-libp2p-pubsub implementation closely

Depends on https://github.com/libp2p/js-libp2p/pull/683
As noted in a comment there, we now require the `libp2p` repo as a full-fledged dependency, due to needing to `openAndCertify` the raw envelope.


wemeetagain:
@vasco-santos  This is ready for re-review. Codecov is giving an :x: but we'll tackle the tests in another PR (in progress :smiley:)

ts/index.ts
```diff
@@ -1225,9 +1309,26 @@ class Gossipsub extends BasicPubsub {
...
     const backoff = constants.GossipsubPruneBackoff / 1000
+    const px: PeerInfo[] = []
+    if (doPX) {
+      // select peers for Peer eXchange
+      const peers = getGossipPeers(this, topic, constants.GossipsubPrunePeers, (xid: string): boolean => {
+        return xid !== id && this.score.score(xid) >= 0
+      })
+      peers.forEach(p => {
+        // see if we have a signed record to send back; if we don't, just send
+        // the peer ID and let the pruned peer find them in the DHT -- we can't trust
```
vasco-santos, line 1320: should we add something here like: "not supported at the time of writing in js-libp2p"?

PR #101, opened 2020-07-02
gs1.1 libp2p param

wemeetagain:
- Replace gossipsub `peerId`, `registrar`, `connectionManager`, etc. params with a `Libp2p` object.
- Fix all tests to suit
  - minor test cleanup where needed
  - ws connections in node, relayed connections in browser see https://github.com/ChainSafe/js-libp2p-gossipsub/pull/99#issuecomment-652511107


test/floodsub.spec.js
```diff
@@ -7,54 +7,46 @@ chai.use(require('dirty-chai'))
...
-  let registrarRecords = Array.from({ length: 2 })
-
   describe('basics', () => {
     let nodeGs
     let nodeFs
 
     beforeEach(async () => {
-      registrarRecords[0] = {}
-      registrarRecords[1] = {}
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
```
vasco-santos, line 30: createPeer currently does not have the started option.

vasco-santos: There are other occurrences. I will not point them out

test/utils/createPeer.js
```diff
@@ -0,0 +1,119 @@
...
+    pubsub: {
+      enabled: false
+    },
+    peerDiscovery: {
+      autoDial: false
+    }
+  }
+}
+
+function isBrowser() {
```
vasco-santos, line 38: Should we use https://github.com/ipfs/js-ipfs-utils/blob/master/src/env.js#L6 here?

test/2-nodes.spec.js
```diff
@@ -51,26 +51,22 @@ describe('1 node', () => {
...
-      const onConnect1 = registrarRecords[1][multicodec].onConnect
+    after(() => Promise.all(nodes.map(stopNode)))
 
-      // Notice peers of connection
-      const [d0, d1] = ConnectionPair()
-      onConnect0(nodes[1].peerId, d0)
-      onConnect1(nodes[0].peerId, d1)
+    it('Dial from nodeA to nodeB happened with pubsub', async () => {
+      await nodes[0]._libp2p.dialProtocol(nodes[1]._libp2p.peerId, multicodec)
+      await new Promise((resolve) => setTimeout(resolve, 10))
```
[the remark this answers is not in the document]
wemeetagain: It seems we do for reliably passing. perhaps there is something not awaited somewhere inside.
Will use `delay` throughout for this.

test/2-nodes.spec.js
```diff
@@ -43,6 +42,7 @@ describe('1 node', () => {
     })
 
     it('should start a gossipsub successfully', async () => {
+      await gossipsub._libp2p.start()
```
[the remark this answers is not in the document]
wemeetagain: :+1:  Will add `started` option to `createPeer`, use throughout where appropriate

test/utils/createPeer.js
wemeetagain: :+1: this is just my hacked attempt, will use the preferred, more robust tool

wemeetagain: Turns out it doesn't detect webworkers. Keeping my version for now.

wemeetagain: lol, I didn't read the file, they separate isBrowser from isWebworker, will add it

wemeetagain: :thinking:  for some reason, using it is causing problems. `isWebWorker`is not working as intended, and `isNode` may not be working? Its causing errors on the node side, using `isNode`.

I think I'd prefer to punt on this, fix in `libp2p-test`

vasco-santos: wow, just doing `!isNode` does not get that working?

vasco-santos: if it does not work, we can punt yes. But that is strange because we are using that across all IPFS land

wemeetagain: it works, in some respect, but causes tests in node fail, with unhandled promise rejection, very strange. Specifically the floodsub.spec.js 'subscription functionality' 'after all' hook. Didn't look into it much more than that.

vasco-santos: ok, lets leave it for now

PR #102, opened 2020-07-02
gs1.1: use publishThreshold for floodsub fallback peers

wemeetagain:
We should check a peer's score against the publishThreshold even if they are a floodsub peer.
This was missed in integrating the scoring into gossipsub.

Caught looking at go-libp2p-pubsub implementaiton


wemeetagain:
@vasco-santos bump
This is just a small PR fixing something I'd missed in the scoring integration

PR #104, opened 2020-07-02
gs1.1: feat: track ihave/iwant counts

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#spam-protection-measures IWANT and IHAVE limiting

Closely follows go-libp2p-pubsub impl

### DIFF
--- a/ts/heartbeat.ts
+++ b/ts/heartbeat.ts
@@ -90,6 +90,10 @@ export class Heartbeat {
     // clean up expired backoffs
     this.gossipsub._clearBackoff()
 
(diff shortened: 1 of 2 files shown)


wemeetagain:
Yeah, would prefer to do as a followup. I'm planning on just copying the go-libp2p-pubsub tests entirely once we have all the functionality.

wemeetagain:
can you take a look at #102 too when you get a chance? small 2 line diff

Issue #105, opened 2020-07-03
Trouble attributing validation error


wemeetagain:
Gossipsubv1.1 has the following spam protection:

> In flight IWANT requests, sent as a response to an IHAVE advertisement, are probabilistically tracked. For each IHAVE advertisement which elicits an IWANT request, the router tracks a random message ID within the advertised set. If the message is not received (from any peer) within a period of time, then a behavioural penalty is applied to the advertising peer through P₇. This measure helps protect against spam IHAVE floods by quickly flagging and graylisting peers who advertise bogus message IDs and/or do not follow up to the IWANT requests.

In go-libp2p-pubsub, they can attribute the invalid message to a specific reason. This lets them properly score the peer/message based on the reason: https://github.com/libp2p/go-libp2p-pubsub/blob/master/score.go#L563

Unfortunately, based on the current design of js-libp2p-pubsub, we are not able to determine _why_ a message is invalid. Only that it is (it returns a boolean, valid/invalid). Earlier in the gossipsubv1.1 effort, we extended our extension of `validate` to include topic validators and a function to process topic validator errors: [here](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/gsv1.1/ts/pubsub.js#L216), but this is insufficient to track other causes of a message being invalid, eg: invalid or missing message signature.

A fix for this very likely requires modifying js-libp2p-pubsub in some way.

One possibility is to have `validate` throw an `Error` using [`js-errcode`](https://github.com/IndigoUnited/js-err-code).
We could then extend `validate` to handle errors of different codes, and make our scoring code quite similar to the above go-libp2p-pubsub example.

Here is another relevant section of go-libp2p-pubsub: https://github.com/libp2p/go-libp2p-pubsub/blob/8945f91465bc3ba418688684a2c29c384cd79647/pubsub.go#L924

Open to other ideas of how to proceed.

wemeetagain:
@vasco-santos

wemeetagain:
resolved in #110

PR #107, opened 2020-07-06
feat: add iwant request tracking

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#spam-protection-measures bullet point 4

Follows go-libp2p-pubsub implementation closely

Suffers from issues related to #105 
eg: in cases where a peer has signaled that they IHAVE a message, we send a IWANT, they send the message, but don't include a signature on a message when we have strict signing, the peer should be penalized twice, once for the invalid message, and once more if they fail to deliver on their promise to send the IWANTed message.
Right now, they only get the first penalty, from the invalid message.

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -14,6 +14,7 @@ import { getGossipPeers } from './getGossipPeers'
 import { createGossipRpc, shuffle, hasGossipProtocol } from './utils'
 import { Peer } from './peer'
 import { PeerScore, PeerScoreParams, PeerScoreThresholds, createPeerScoreParams, createPeerScoreThresholds } from './score'
+import { IWantTracer } from './tracer'
 import { AddrInfo, Libp2p } from './interfaces'
 // @ts-ignore
 import TimeCache = require('time-cache')
@@ -52,6 +53,7 @@ class Gossipsub extends BasicPubsub {
   outbound: Map<Peer, boolean>
   score: PeerScore
   heartbeatTicks: number
+  gossipTracer: IWantTracer
   _libp2p: Libp2p
   _options: GossipOptions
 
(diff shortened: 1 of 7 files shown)


ts/tracer.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.promises.delete(msgId)
+      }
+    })
+
+    return result
+  }
+
+  /**
+   * Someone delivered a message, stop tracking promises for it
+   * @param {string} p peer id
```
vasco-santos, line 75: It seems that we are not using the peerID

ts/index.ts
```diff
@@ -671,6 +685,17 @@ class Gossipsub extends BasicPubsub {
     }
   }
 
+  /**
+   * Apply penalties from broken IHAVE/IWANT promises
+   * @returns {void}
+   */
+  _applyIwantPenalties (): void {
```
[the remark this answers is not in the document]
wemeetagain: ah! thanks, should be added to _heatbeat

ts/tracer.ts
wemeetagain: The idea is to have the same function signature as `PeerScore#deliverMessage`.
So that we might in the future abstract the different 'tracers' as in go-libp2p-pubsub
See https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L972
Eg: they don't call each one individually:
```
this.score.deliverMessage(p, msg)
this.gossipTracer.deliverMessage(p, msg)
this.tagTracer.deliverMessage(p, msg)
```
They have some sort of registration for each of their tracers, then all tracers get called on a single call to `deliverMessage`.
Note the go-libp2p-pubsub tracer functionality is happening on the pubsub layer, not the gossipsub layer. Aggregating our "tracers" together would be a first step towards moving in that direction. Not saying thats what we _should_ do, but it is a prerequisite.

vasco-santos: ok, sounds good

PR #108, opened 2020-07-06
chore: update gossipsub multicodec

wemeetagain:


ts/index.ts
```diff
@@ -75,7 +75,7 @@ class Gossipsub extends BasicPubsub {
     libp2p: Libp2p,
     options: Partial<GossipInputOptions> = {}
   ) {
-    const multicodecs = [constants.GossipsubIDv10]
```
vasco-santos, line 78: Shouldn't we have both here? We will want to connect to peers using the `v1.0` protocol.
It would also be important to have a test running each protocol

wemeetagain: :+1: gotcha. Likely we'll need to wait for https://github.com/libp2p/js-libp2p-pubsub/issues/54 for reliable test
Will add v1.0 protocol here

PR #109, opened 2020-07-07
fix: connected with a subscriber before a mesh is created should send

vasco-santos:
@jacobheun discovered this bug where if we're connected to a subscriber before the mesh is formed, we don't send them messages.


vasco-santos:
I am cutting a release on this

wemeetagain:
thanks

PR #121, opened 2020-08-16
gs1.1: chore: fix various bugs

wemeetagain:
~~Depends on #120~~

Several significant issues:
- `!iwant || !ihave || !prune`
  - we were never:
    - responding to ihave messages (with iwants)
    - responding to iwant messages (with actual messages)
    - sending prune messages when we overgrafted with a peer
  - This issue has been present since the birth of this library. In a very real sense, it has always been severely broken.
- shortcircuit ihave/iwant handling
  - in the `handleIWant` case, this is just an optimization to avoid calculating a peer score when we don't need to
  - in the `handleIHave` case, it also does this, but more importantly, stops flood protection from being erroneously triggered on an empty control message. This would cause ihaves to be mosty ignored in practice, as every rpc would count towards ihave tracking.
- call `buildMessage` early in `_publish`
  - This lets us cache the message with a signature attached. Important for sending gossip, we will want the gossip to include the signature if it was included in the original message.

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -352,11 +352,11 @@ class Gossipsub extends BasicPubsub {
     const prune = this._handleGraft(id, controlMsg.graft)
     this._handlePrune(id, controlMsg.prune)
 
-    if (!iwant || !ihave || !prune) {
+    if (!iwant && !ihave && !prune) {
       return
     }
 
-    const outRpc = createGossipRpc(ihave, { iwant: [iwant], prune })
+    const outRpc = createGossipRpc(ihave, { iwant: iwant ? [iwant] : [], prune })
     this._sendRpc(id, outRpc)
   }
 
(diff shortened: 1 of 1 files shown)


ts/index.ts
```diff
@@ -352,11 +352,11 @@ class Gossipsub extends BasicPubsub {
     const prune = this._handleGraft(id, controlMsg.graft)
     this._handlePrune(id, controlMsg.prune)
 
-    if (!iwant || !ihave || !prune) {
+    if (!iwant && !ihave && !prune) {
       return
     }
 
-    const outRpc = createGossipRpc(ihave, { iwant: [iwant], prune })
+    const outRpc = createGossipRpc(ihave, { iwant: iwant ? [iwant] : [], prune })
```
vasco-santos, line 359:
```suggestion
    const outRpc = createGossipRpc(ihave, { iwant: iwant || [], prune })
```

perhaps we can simplify?

PR #122, opened 2020-08-17
chore: use pubsub interface

vasco-santos:
This PR moves away from `libp2p-pubsub` base implementation and uses the new (still in PR) `interface-pubsub`.

The `interface-pubsub` provides some advantages over `libp2p-pubsub`:
- common test suite used by pubsub routers, decreasing tests duplication and future needs in new routers
- guarantees that the pubsub routers have the expected API and behaviour (no need to have a pubsub abstraction in `js-libp2p` per https://github.com/libp2p/js-libp2p-pubsub/issues/68)
- more info https://github.com/libp2p/js-libp2p-interfaces/issues/53

The current base pubsub implementation in the interface is mostly the content of https://github.com/libp2p/js-libp2p-pubsub/tree/v0.6.x with some changes in the constructor.

This PR aims to:

- Use `interface-pubsub` with its base implementation and test suite
- Remove unnecessary tests
- Heartbeat start/stop errors removed
- Handler for subscribe/unsubscribe feature added (remove libp2p abstraction that mostly adds this)

Note: The interface still needs to be updated with the content of `pubsub.js` in this repo.

Needs:

- [ ] https://github.com/libp2p/js-libp2p-interfaces/pull/60
- [ ] https://github.com/libp2p/js-libp2p-floodsub/pull/108


ts/index.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 // @ts-ignore
-import { utils } from 'libp2p-pubsub'
+import { utils } from 'libp2p-pubsub' // TODO: change
```
vasco-santos, line 3: I would like some help on this. I tried to change this line like I did for the others, but ts did not like it

package.json
```diff
@@ -43,6 +43,7 @@
     "denque": "^1.4.1",
     "err-code": "^2.0.0",
     "it-pipe": "^1.0.1",
+    "libp2p-interfaces": "libp2p/js-libp2p-interfaces#feat/interface-pubsub",
     "libp2p-pubsub": "https://github.com/libp2p/js-libp2p-pubsub#v0.6.x",
```
vasco-santos, line 47: This needs to be removed! See issue below

ts/pubsub.js
```diff
@@ -284,9 +283,10 @@ class BasicPubSub extends Pubsub {
    * Subscribes to topics
    * @override
    * @param {Array<string>|string} topics
+   * @param {function} [handler]
    * @returns {void}
    */
-  subscribe (topics) {
+  subscribe (topics, handler) {
```
wemeetagain, line 289: Why attaching an event handler here?

ts/index.ts
wemeetagain: resolved

ts/pubsub.js
vasco-santos: we want to remove this abstraction file completely from libp2p (https://github.com/libp2p/js-libp2p/blob/master/src/pubsub.js). The only thing we really needed was the handlers, as all the rest was basically protection from the APIs per https://github.com/libp2p/js-libp2p-pubsub/issues/68

I am still thinking if we can move that to the interface, but with the subscriptions I am not sure we will be able to abstract it there

wemeetagain: :+1:

RE: moving it to the interface:
We could do that depending on what we 'allow' to be overridable.
It seems that we only need to have routers override `_subscribe(topics: string[]): void` and `_unsubscribe(topics: string[]): void` (which MUST call super). And not `subscribe`/`unsubscribe`
In which case we could perform the event handling on `subscribe` and `unsubscribe` (before/after `_subscribe`/`_unsubscribe` is called).

wemeetagain: Unrelated, but I think the interface should only allow subscription/unsubscription of a single topic at a time.
eg: `subscribe(topic: string, handler: (msg: InMessage) => void): void`

Attaching a handler to multiple topics at once feels a lot like an anti-pattern, or at least a less primary interface, in the same way we were previously publishing multiple topics and multiple messages at once.

vasco-santos: That is a good point. Please create that discussion in the interface PR

PR #123, opened 2020-08-17
chore: add gossip retransmission counter

wemeetagain:
Implements https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#spam-protection-measures bullet 2

Follows go-libp2p-gossipsub implementation closely

### DIFF
--- a/ts/messageCache.ts
+++ b/ts/messageCache.ts
@@ -7,6 +7,7 @@ export interface CacheEntry {
 
 export class MessageCache {
   msgs: Map<string, InMessage>
+  peertx: Map<string, Map<string, number>>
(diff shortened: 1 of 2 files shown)


ts/messageCache.ts
```diff
@@ -74,6 +77,32 @@ export class MessageCache {
     return this.msgs.get(msgID)
   }
 
+  /**
+   * Retrieves a message from the cahche by its ID, if it is present
```
vasco-santos, line 81:
```suggestion
   * Retrieves a message from the cache by its ID, if it is present
```

PR #125, opened 2020-08-18
chore: consistent files naming

vasco-santos:
As previously talked, changed all the files named in camelCase to dash-case, in order to be consistent inside all this repo, as well as in the libp2p modules


wemeetagain:
file conflicts after merging other PRs

vasco-santos:
I will have a look at it

PR #130, opened 2020-08-25
chore: use libp2p-interfaces0.5.1 and libp2p0.29 rc

vasco-santos:
Updates libp2p interfaces and floodsub

Removed floodsub uneeded tests with multiple messages as they are included in the interface for floodsub


vasco-santos:
@wemeetagain I will also add to this PR the libp2p rc so that we can release gossipsub

PR #140, opened 2020-09-17
Libp2p interfaces types

mpetrunic:


ts/index.ts
```diff
@@ -319,7 +318,7 @@ class Gossipsub extends Pubsub {
         }
       }
     }
-    this.outbound.set(p, outbound)
+    this.outbound.set(p.id.toB58String(), outbound)
```
mpetrunic, line 321: @tuyennhv @wemeetagain Is this ok? types aren't matching since map is string => boolean, while p is PeerStreams

package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "lint": "eslint --ext .ts ts",
     "release": "aegir release",
-    "prebuild": "tsc",
+    "prebuild": "tsc --traceResolution",
```
tuyennhv, line 15: we don't need this

ts/index.ts
tuyennhv: this looks good to me, I see `this.outbound` works with `peerId.toB58String` as key in other places

package.json
```diff
@@ -34,16 +34,13 @@
...
-    "lint"
-  ],
   "dependencies": {
     "@types/debug": "^4.1.5",
     "debug": "^4.1.1",
     "denque": "^1.4.1",
     "err-code": "^2.0.0",
     "it-pipe": "^1.0.1",
-    "libp2p-interfaces": "^0.5.1",
+    "libp2p-interfaces": "https://github.com/mpetrunic/js-libp2p-interfaces.git#mpetrunic/typescript-types",
```
tuyennhv, line 43: libp2p-interface 0.5.2 is ready, let's use it

tuyennhv:
not sure how to get through build error

ts/index.ts
```diff
@@ -1,28 +1,27 @@
-/* eslint-disable @typescript-eslint/ban-ts-comment */
-// @ts-ignore
-import { utils } from 'libp2p-interfaces/src/pubsub'
+import Pubsub, { InMessage, utils } from 'libp2p-interfaces/src/pubsub'
 import { MessageCache } from './message-cache'
 import {
   RPCCodec,
-  RPC, Message, InMessage,
+  RPC, Message,
```
wemeetagain, line 5: Might be good to move `Message` to libp2p-interfaces too.
Future PR

mpetrunic: Argh, we already have that in libp2p-interfaces just need to change import, but we can do that in separate PR if this PR is blocker for gossipsub migration?

tuyennhv (approved):

.travis.yml
```diff
@@ -6,7 +6,7 @@ stages:
   - cov
 
 node_js:
-  - '10'
+  - '12'
```
vasco-santos, line 9: I would recommend we do:

```
node_js:
  - 'lts/*'
  - 'stable'
```

We can do a follow up PR for this!

ts/tracer.ts
```diff
@@ -92,6 +90,7 @@ export class IWantTracer {
    * A message got rejected, so we can stop tracking promises and let the score penalty apply from invalid message delivery,
    * unless its an obviously invalid message.
    * @param {InMessage} msg
+   * @param reason
```
vasco-santos, line 93:
```suggestion
   * @param {string} reason
```

.travis.yml
mpetrunic: libp2p doesn't allow node12 anymore:
> error libp2p@0.29.0: The engine "node" is incompatible with this module. Expected version ">=12.0.0". Got "10.17.0"

vasco-santos (approved):

PR #141, opened 2020-09-20
Add message signing options

tuyennhv:
## Goal
+ For eth2, we want to disable message signing options `{signMessages: false, strictSigning: false}`
+ libp2p-interface allows us to do that, refer to https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/pubsub/index.js#L49

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,8 @@ Options is an optional object with the following key-value pairs:
 * **`floodPublish`**: boolean identifying if self-published messages should be sent to all peers, (defaults to **true**).
 * **`doPX`**: boolean identifying whether PX is enabled; this should be enabled in bootstrappers and other well connected/trusted nodes (defaults to **false**).
 * **`msgIdFn`**: a function with signature `(message) => string` defining the message id given a message, used internally to deduplicate gossip (defaults to `(message) => message.from + message.seqno.toString('hex')`)
+* **`signMessages`**: boolean identifying if we want to sign outgoing messages or not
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -62,6 +62,8 @@ Options is an optional object with the following key-value pairs:
 * **`floodPublish`**: boolean identifying if self-published messages should be sent to all peers, (defaults to **true**).
 * **`doPX`**: boolean identifying whether PX is enabled; this should be enabled in bootstrappers and other well connected/trusted nodes (defaults to **false**).
 * **`msgIdFn`**: a function with signature `(message) => string` defining the message id given a message, used internally to deduplicate gossip (defaults to `(message) => message.from + message.seqno.toString('hex')`)
+* **`signMessages`**: boolean identifying if we want to sign outgoing messages or not
```
vasco-santos, line 65:
```suggestion
* **`signMessages`**: boolean identifying if we want to sign outgoing messages or not (default: `true`)
```

README.md
```diff
@@ -62,6 +62,8 @@ Options is an optional object with the following key-value pairs:
...
+* **`strictSigning`**: boolean identifying if message signing is required for incoming messages or not
```
vasco-santos, line 66:
```suggestion
* **`strictSigning`**: boolean identifying if message signing is required for incoming messages or not (default: `true`)
```

vasco-santos (changes requested):
This looks good, just a suggestion on the docs default

vasco-santos (approved):
LGTM!

PR #143, opened 2020-10-10
Support Uint8Array message id

wemeetagain:
Following libp2p/js-libp2p-interfaces#65 , support Uint8Array message IDs

### DIFF
--- a/test/message-cache.spec.js
+++ b/test/message-cache.spec.js
@@ -12,13 +12,10 @@ const uint8ArrayFromString = require('uint8arrays/from-string')
 
 const { MessageCache } = require('../src/message-cache')
 const { utils } = require('libp2p-interfaces/src/pubsub')
-
-const getMsgID = (msg) => {
-  return utils.msgId(msg.from, msg.seqno)
-}
+const { getMsgId } = require('./utils')
 
 describe('Testing Message Cache Operations', () => {
-  const messageCache = new MessageCache(3, 5, getMsgID)
+  const messageCache = new MessageCache(3, 5, getMsgId)
   const testMessages = []
 
   before(() => {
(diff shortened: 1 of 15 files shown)


test/message-cache.spec.js
```diff
@@ -42,8 +39,8 @@ describe('Testing Message Cache Operations', () => {
 
   it('Should retrieve correct messages for each test message', () => {
     for (let i = 0; i < 10; i++) {
-      const msgId = getMsgID(testMessages[i])
```
dapplion, line 45: Rename variable to `messageID` for consistency

test/utils/index.js
```diff
@@ -3,6 +3,7 @@
 const { expect } = require('chai')
 
 const FloodSub = require('libp2p-floodsub')
+const { result } = require('lodash')
```
dapplion, line 6: Typo, remove

dapplion (changes requested):
minor typos

ts/utils/messageIdToString.ts
```diff
@@ -0,0 +1,3 @@
+export function messageIdToString (msgId: Uint8Array): string {
+  return (new Uint8Array(msgId)).toString()
```
vasco-santos, line 2: Use https://github.com/achingbrain/uint8arrays#tostringarray-encoding--utf8

wemeetagain: :+1: I guess we can use base16 or base64? The msg id won't be guaranteed to be valid utf8

dapplion: Is it not safe to use `.toString()` directly? While the resulting string is not good for readability it may be good enough for comparison

vasco-santos: @wemeetagain we can use `base64` 👍 
@dapplion if we need to revert it back to Uint8Array anytime in the future, it will cause issues.

wemeetagain: I wish that library had type definitions :sweat_smile:

wemeetagain:
ready for re-review

3xtr4t3rr3str14l (approved):

dapplion (approved):
:fire:

vasco-santos (approved):
LGTM

PR #144, opened 2020-11-07
Update README.md

DanPen:
`emitSelf` appears to default to `true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ const gossipsub = new Gossipsub(libp2p, options)
 
 Options is an optional object with the following key-value pairs:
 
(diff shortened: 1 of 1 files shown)


vasco-santos (commented):
Hey @DanPen 
The `emitSelf` should be false by default. If you check the [pubsub interface](https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/pubsub/index.js#L59), you can see it. On `libp2p`, the default is overwritten on its [config](https://github.com/libp2p/js-libp2p/blob/v0.29.3/src/config.js#L53). Maybe this where the confusion appeared?
We are keeping the `emitSelf` to be true on libp2p to avoid the breaking change, but we should move on that way.

vasco-santos:
The emitSelf default was changed on libp2p. Going to close this PR

PR #146, opened 2020-12-02
chore: update to new libp2p interface with types

vasco-santos:
`libp2p-interfaces` now supports via aegir types generation on release. This PR updates gossipsub to use it and fixes a couple of ts issues.

At the moment, it is using a branch from the PR below that has a `prepare` script to generate the build while we do not release it.

Needs:

- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/74
- [x] Re-add src from .gitignore
- [x] libp2p@0.30


vasco-santos:
cc @wemeetagain 
This should be ready for review, but somehow CI has been broken in this PR/repo. Any ideas? Cannot access travis from here

tsconfig.json
```diff
@@ -2,19 +2,20 @@
...
     "lib": [
-      "es6"
+      "es6", "dom"
     ],
     "target": "es6",
     "noImplicitAny": true,
     "noImplicitThis": true,
     "strictFunctionTypes": true,
     "strictNullChecks": true,
+    "skipLibCheck": true,
```
vasco-santos, line 12: It seems that `typescript@4.1.3` released yesterday broke a type def... Still need to check, we have been experiencing a lot of issues with typescript releases changing outputs of ts declaration files generated.

wemeetagain (approved):
:raised_hands:

PR #153, opened 2021-04-07
chore: update deps and remove protons

vasco-santos:
Updating all the dependencies and removing protons in favour of protobufjs for smaller bundles and avoiding bundle problems with web-encoding. This follows the work that is in progress in libp2p to use protobufjs everywhere. As a bonus, we can create type definitions automatically.

`rpc.d.ts` and `rpc.js` were generated by protobufjs CLI based on the proto file. Scripts for generating/updating them were also added

Needs:
- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/89


vasco-santos:
@hugomrdias I am having issues with mocha in playwright similar to https://github.com/hugomrdias/playwright-test/issues/52

Are you aware of any regression?

vasco-santos:
We need to update gossipsub build step to rely on aegir. We can use noise as an example. I don't want to add more scope to this PR, so I will create an issue to track this change

scripts/node-globals.js
```diff
@@ -0,0 +1 @@
+export const { Buffer } = require('buffer')
```
rvagg, line 1: ditto for `"buffer"` dependency - I'm assuming we're supporting browser builds here too

.github/workflows/main.yml
```diff
@@ -12,10 +12,10 @@ jobs:
...
-    - run: yarn lint
-    - run: yarn prebuild
-    - run: yarn aegir dep-check
+    - run: npm install
+    - run: npm run lint
+    - run: npm run prebuild
+    - run: npx aegir dep-check
   test-node:
     needs: check
     runs-on: ${{ matrix.os }}
```
rvagg, line 21: `node` version below here needs updating

rvagg (dismissed):
sgtm, lots of `?` + `|null` again but I guess that's reasonable for now, can improve later
(also pending dev dep)

vasco-santos:
> sgtm, lots of ? + |null again but I guess that's reasonable for now, can improve later

 https://github.com/libp2p/js-libp2p-interfaces/pull/89#issuecomment-817339489

vasco-santos:
This PR is having some CI troubes, as it has some old hanging jobs

wemeetagain (approved):
LGTM :tada:

achingbrain:
@vasco-santos the `prepare` script need to generate the protobuf files too, otherwise you get:

```
../../node_modules/libp2p-gossipsub/src/index.js:32:22: error: Could not resolve "./message/rpc"
  32 │ const rpc_1 = require("./message/rpc");
```

```
$ ls -l node_modules/libp2p-gossipsub/src
total 264
-rw-r--r--   1 alex  staff   8304 26 Oct  1985 constants.d.ts
-rw-r--r--   1 alex  staff   9237 26 Oct  1985 constants.js
-rw-r--r--   1 alex  staff    472 26 Oct  1985 get-gossip-peers.d.ts
-rw-r--r--   1 alex  staff   1235 26 Oct  1985 get-gossip-peers.js
-rw-r--r--   1 alex  staff    702 26 Oct  1985 heartbeat.d.ts
-rw-r--r--   1 alex  staff  13885 26 Oct  1985 heartbeat.js
-rw-r--r--   1 alex  staff  11413 26 Oct  1985 index.d.ts
-rw-r--r--   1 alex  staff  44334 26 Oct  1985 index.js
-rw-r--r--   1 alex  staff    275 26 Oct  1985 interfaces.d.ts
-rw-r--r--   1 alex  staff     77 26 Oct  1985 interfaces.js
-rw-r--r--   1 alex  staff   1929 26 Oct  1985 message-cache.d.ts
-rw-r--r--   1 alex  staff   3522 26 Oct  1985 message-cache.js
drwxr-xr-x  20 alex  staff    640 16 Apr 11:18 score
-rw-r--r--   1 alex  staff   1739 26 Oct  1985 tracer.d.ts
-rw-r--r--   1 alex  staff   3356 26 Oct  1985 tracer.js
drwxr-xr-x  12 alex  staff    384 16 Apr 11:18 utils
```

vasco-santos:
bah, not generating because the file is commited. But it needs to be moved to src on ts compile. I will change it in a sec

vasco-santos:
@achingbrain done, can you retry?

achingbrain:
@vasco-santos Much better, thanks!

vasco-santos:
Prior to https://github.com/libp2p/js-libp2p/commit/b4fb9b7bf266ba03c4462c0a41b1c2691e4e88d4 the event was emitted with empty addresses, which meant the findPeer was working, but no addresses existed. In addition, test dht connect now also adds to the addressBook the peer addresses onConnect, as libp2p would do if we were using its dialer

PR #155, opened 2021-07-09
chore: update deps

achingbrain:
There's a new release of the `peer-id` module that uses the `CID` class from the `multiformats` module instead of from `cids`.

This PR upgrades to the new version, along with all the supporting ecosystem modules

BREAKING CHANGE: the new `peer-id` module uses the new `CID` class and not the old one


achingbrain:
A draft PR as it needs https://github.com/libp2p/js-libp2p-floodsub/pull/116 to land in order to remove the final gh dep url

achingbrain:
Ha! https://github.com/libp2p/js-libp2p-floodsub/pull/116 landed and the branch was deleted making CI sad.  One moment caller.

vasco-santos (approved):

PR #156, opened 2021-07-16
chore: use latest libp2p release

vasco-santos:
It is important pointing out that `getMsgId` function might return a Promise per https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interfaces/src/pubsub/index.js#L479 We were having issues in browser tests thanks to it, so I needed to make it accept promises too.

Also updates uint8arrays

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
     "it-pair": "^1.0.0",
-    "libp2p": "next",
+    "libp2p": "^0.32.0",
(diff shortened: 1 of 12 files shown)


achingbrain:
@wemeetagain @mpetrunic it would be really helpful if this could be merged & released

dapplion:
> @wemeetagain @mpetrunic it would be really helpful if this could be merged & released

Is this safe to merge as is? I can trigger the release now

vasco-santos:
> Is this safe to merge as is? I can trigger the release now

Yes, it is good to go

dapplion (approved):

PR #159, opened 2021-09-03
fix: remove yarn.lock

vasco-santos:
This PR removes yarn.lock as it seems accidentally added, it was also ignored for avoid future similar accidents


vasco-santos:
My confusion, we do not have a package-lock here, we just ignore package-lock, but not yarn lock. Closing for now

dapplion (approved):

PR #165, opened 2021-09-13
feat: add more configuration options

wemeetagain:
There are many options currently set by constants.
This is limiting when applications want to override these for
application-specific networks.

Several interesting options are now overridable via the constructor:

- heartbeatInterval
- fanoutTTL
- mcacheLength
- mcacheGossip
- seenTTL

These options are optional with fallback values defaulting to the prior "constant" values.


dapplion (approved):

vasco-santos (approved):
LGTM

PR #167, opened 2021-10-04
Fix pubsub interface link

D4nte:
As it points to a deprecated repo which itself points to a broken link.


vasco-santos (approved):
thanks

PR #172, opened 2021-12-01
Implement SimpleTimeCache

### LINKED ISSUE #168: TimeCache performance issue
**Motivation**
`TimeCache.put()` takes almost 3% of the total time of lodestar application, we need to consider implementing our own cache instead.
Screen Shot 2021-11-30 at 13 08 32

[1130_lodestar_subscribeAllSubnets_local.cpuprofile.zip](https://github.com/ChainSafe/js-libp2p-gossipsub/files/7623334/1130_lodestar_subscribeAllSubnets_local.cpuprofile.zip)

tuyennhv:
**Motivation**
+ Need to implement a faster time cache

**Description**
+ Implement SimpleTimeCache
+ The benchmark test shows that it's 4x faster than the current npm TimeCache

Close #168

### DIFF
--- a/ts/utils/time-cache.ts
+++ b/ts/utils/time-cache.ts
@@ -1,0 +1,49 @@
+type SimpleTimeCacheOpts = {
+  validityMs: number
+}
+
+/**
+ * This is similar to https://github.com/daviddias/time-cache/blob/master/src/index.js
+ * for our own need, we don't use lodash throttle to improve performance.
+ * This gives 4x - 5x performance gain compared to npm TimeCache
+ */
+export class SimpleTimeCache {
+  private entries: Map<string, number>;
+  private validityMs: number;
+  private lastPruneTime = 0;
+
+  constructor (options: SimpleTimeCacheOpts) {
+    this.entries = new Map()
+    this.validityMs = options.validityMs
+
+    if (this.validityMs <= 0) {
+      throw Error(`Invalid validityMs ${this.validityMs}`)
+    }
+  }
+
+  put (key: string): void {
+    this.entries.set(key, Date.now())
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -1,7 +1,6 @@
 .idea/
 node_modules/
 src/
-package-lock.json
```
dapplion, line 4: Don't commit `package-lock.json` only `yarn.lock`

ts/utils/time-cache.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.entries.set(key, Date.now())
+    this.prune()
+  }
+
+  prune (): void {
+    const now = Date.now()
+    if (now - this.lastPruneTime < 200) {
+      return
+    }
+    this.lastPruneTime = now
```
dapplion, line 34: So much simpler and elegant :+1: :tada:

.gitignore
tuyennhv: @wemeetagain @dapplion we use npm in all of our github workflows, and we instruct to use it in README as well. Do we really want to switch to yarn?

Also I have to use node 14 and freeze dependencies to get through CI so I have to commit `package-lock.json`

dapplion: @wemeetagain Can you take a look?

dapplion: @tuyennhv please stick to yarn.lock in this PR and open an issue about switching to package-lock.json. This PR can't be +14000 lines and we can discuss this latter

tuyennhv: @dapplion I just rebase to master, we don't need to commit `package-lock.json` anymore

dapplion (approved):

PR #173, opened 2021-12-02
chore: update peer-id

achingbrain:
Pulls in the new peer-id

Needs a new release of `@chainsafe/libp2p-noise` though it's just a dev dep so this could go out without it.

BREAKING CHANGE: requires node 15+


dapplion:
@achingbrain what breaks with Node < v16?

achingbrain:
Node only used to be able to give us keypairs in pem format, which we'd then use [pem-jwk](https://npmjs.com/package/pem-jwk) to transform to jwk - starting with node 15 it can give us the keys in jwk directly so we don't need the extra dep any more.

https://github.com/libp2p/js-libp2p-crypto/blob/master/src/keys/rsa.js#L14-L15

achingbrain:
Same with signing/verifying using RSA, we used to have to transform the key format before hand, now we can just pass it to node crypto directly.

dapplion:
> Node only used to be able to give us keypairs in pem format, which we'd then use [pem-jwk](https://npmjs.com/package/pem-jwk) to transform to jwk - starting with node 15 it can give us the keys in jwk directly so we don't need the extra dep any more.
> 
> https://github.com/libp2p/js-libp2p-crypto/blob/master/src/keys/rsa.js#L14-L15

Is it feasible to have a fallback for Node v14? I haven't though in depth our node version support strategy but if the cost is low it's a good to have.

achingbrain:
It means adding the `pem-jwk` dep back in and not using the features from node 15.  Node 16 is Active LTS now though, so that's the minimum version we support, [officially](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#supported-versions).

dapplion:
> It means adding the `pem-jwk` dep back in and not using the features from node 15. Node 16 is Active LTS now though, so that's the minimum version we support, [officially](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#supported-versions).

Agree with your policy, let's stick to v16 then

dapplion (approved):

achingbrain:
Phew, CI passing.  I think long term all the `delay(...)`s need to be removed from the tests, some of the values are too aggressive for resource constrained environments like GH actions.

test/gossip.spec.js
```diff
@@ -31,9 +32,9 @@ describe('gossip', () => {
     // add subscriptions to each node
     nodes.forEach((n) => n.subscribe(topic))
 
+    // every node connected to every other
     await connectGossipsubs(nodes)
-    // await subscription propagation
-    await delay(50)
+    await waitForAllNodesToBePeered(nodes)
```
vasco-santos, line 37: awesome

vasco-santos (approved):
LGTM

PR #175, opened 2021-12-06
Cache peer score

### LINKED ISSUE #169: computeScore performance issue
**Motivation**
+ We keep computing peer score in a lot of places without any cache
+ It's not scalable: the more messages, the more function calls we have for `computeScore()()`. When there are a lot of messages, it takes 4% cpu time of lodestar
[1130_lodestar_subscribeAllSubnets_local.cpuprofile.zip](https://github.com/ChainSafe/js-libp2p-gossipsub/files/7623357/1130_lodestar_subscribeAllSubnets_local.cpuprofile.zip)

Screen Shot 2021-11-30 at 13 14 38

**Note**
+ It's not likely it's a performance issue of `computeScore()` function itself

tuyennhv:
**Motivation**
+ In lodestar, it takes 4% of cpu time just to compute peer score

**Description**
+ Cache peer score if topic stat & peer stat are unchanged for maximum `decayInterval` time period
+ Main consumer of `computeScore()` is `acceptFrom()` function which is called per incoming message. If peer score >= 0 we allow up to 128 messages and 1s not to compute score (this is safe enough considering `grayListThreshold=-16000`
+ Closes #169

Thanks Teku for the work https://github.com/libp2p/jvm-libp2p/pull/184

### DIFF
--- a/test/accept-from.spec.js
+++ b/test/accept-from.spec.js
@@ -1,0 +1,79 @@
+const {expect} = require('chai')
+const sinon = require('sinon')
+const {PeerScore} = require('../src/score')
+const Gossipsub = require('../src')
+const {
+  createPeer,
+} = require('./utils')
+
+describe('Gossipsub acceptFrom', () => {
+  let gossipsub
+  const sandbox = sinon.createSandbox()
+  let scoreStub
+
+  beforeEach(async () => {
+    sandbox.useFakeTimers()
+    gossipsub = new Gossipsub(await createPeer({ started: false }), { emitSelf: true })
+    scoreStub = sandbox.createStubInstance(PeerScore)
+    gossipsub.score = scoreStub
+  })
+
+  afterEach(() => {
+    sandbox.restore()
+  })
+
+  it('should only white list peer with positive score', () => {
+    scoreStub.score.withArgs("peerA").returns(1000)
+    gossipsub._acceptFrom("peerA")
+    // 1st time, we have to compute score
+    expect(scoreStub.score.withArgs("peerA").calledOnce).to.be.true
+    // 2nd time, use a cached score since it's white listed
+    gossipsub._acceptFrom("peerA")
+    expect(scoreStub.score.withArgs("peerA").calledOnce).to.be.true
+  })
+
+  it('should recompute score after 1s', () => {
+    scoreStub.score.returns(1000)
+    gossipsub._acceptFrom("peerA")
+    // 1st time, we have to compute score
+    expect(scoreStub.score.withArgs("peerA").calledOnce).to.be.true
+    gossipsub._acceptFrom("peerA")
+    expect(scoreStub.score.withArgs("peerA").calledOnce).to.be.true
+
+    // after 1s
+    sandbox.clock.tick(1001)
+
+    gossipsub._acceptFrom("peerA")
+    expect(scoreStub.score.withArgs("peerA").calledTwice).to.be.true
+  })
+
+  it('should recompute score after max messages accepted', () => {
+    scoreStub.score.returns(1000)
+    gossipsub._acceptFrom("peerA")
+    // 1st time, we have to compute score
+    expect(scoreStub.score.withArgs("peerA").calledOnce).to.be.true
+
+    for (let i = 0; i < 128; i++) {
+      gossipsub._acceptFrom("peerA")
+    }
+    expect(scoreStub.score.withArgs("peerA").calledOnce).to.be.true
+
+    // max messages reached
+    gossipsub._acceptFrom("peerA")
+    expect(scoreStub.score.withArgs("peerA").calledTwice).to.be.true
+  })
+
+  it('should NOT white list peer with negative score', () => {
+    // peerB is not white listed since score is negative
+    scoreStub.score.withArgs("peerB").returns(-1)
+    gossipsub._acceptFrom("peerB")
+    // 1st time, we have to compute score
+    expect(scoreStub.score.withArgs("peerB").calledOnce).to.be.true
+    // 2nd time, still have to compute score since it's NOT white listed
+    gossipsub._acceptFrom("peerB")
+    expect(scoreStub.score.withArgs("peerB").calledTwice).to.be.true
+  })
+
+
+
+})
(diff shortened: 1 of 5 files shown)


ts/index.ts
```diff
@@ -84,10 +84,22 @@ interface GossipOptions extends GossipInputOptions {
...
+  messagesAccepted: number
+  /** have to recompute score after this time */
+  whitelistedTill: number
+}
+
 class Gossipsub extends Pubsub {
   peers: Map<string, PeerStreams>
   direct: Set<string>
   seenCache: SimpleTimeCache
+  acceptRequestsWhitelist: Map<string, AcceptRequestWhiteListEntry>
```
wemeetagain, line 102: We can call this the `acceptFromWhitelist`. (and replace "request" elsewhere too)

We should add a comment here to explain the purpose, too.

ts/index.ts
```diff
@@ -84,10 +84,22 @@ interface GossipOptions extends GossipInputOptions {
...
 
+const ACCEPT_REQUEST_WHITE_LIST_THRESHOLD_SCORE = 0
+const ACCEPT_REQUEST_WHITE_LIST_MAX_MESSAGES = 128
+const ACCEPT_REQUEST_WHITE_LIST_DURATION_MS = 1000
+
+interface AcceptRequestWhiteListEntry {
+  /** max number of messages accepted after a score is calculated */
+  messagesAccepted: number
+  /** have to recompute score after this time */
+  whitelistedTill: number
```
wemeetagain, line 95: can you call this `acceptUntil`

ts/index.ts
```diff
@@ -84,10 +84,22 @@ interface GossipOptions extends GossipInputOptions {
   scoreThresholds: PeerScoreThresholds
 }
 
+const ACCEPT_REQUEST_WHITE_LIST_THRESHOLD_SCORE = 0
```
wemeetagain, line 87: Can you move these constants to ts/constants and also add jsdoc comments for each?

wemeetagain (changes requested):

ts/score/peer-score.ts
```diff
@@ -30,6 +30,18 @@ export class PeerScore {
...
    */
   peerIPs: Map<string, Set<string>>
+  /**
+   * Cache score up to decayInterval if topic stats are unchanged.
+   */
+  scoreCache: Map<string, number>
+  /**
+   * Flag to mark a peer score cache valid or not.
+   */
+  scoreCacheValid: Map<string, boolean>
```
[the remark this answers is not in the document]
tuyennhv: yeah that's a good idea 👍 , we can save 1 variable. Instead of
```
scoreCacheValid.set(id, false)
```
do
```
scoreCacheTime.delete(id)
```

when calling `score(id)`, check for the existence of `scoreCacheTime` first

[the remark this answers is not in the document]
tuyennhv: Update: I changed from `scoreCacheTime` to `scoreCacheUntil` as the other comment

ts/index.ts
```diff
@@ -182,6 +191,13 @@ class Gossipsub extends Pubsub {
      */
     this.direct = new Set(opts.directPeers.map(p => p.id.toB58String()))
 
+    /**
+     * Map of peer id and AcceptRequestWhileListEntry
+     *
+     * @type {Map<string, AcceptFromWhiteListEntry}
+     */
+    this.acceptFromWhitelist = new Map()
```
wemeetagain, line 199: I think we need a place where these entries get routinely pruned so they don't build up.
Maybe you can prune old entries on every N heartbeats

wemeetagain (changes requested):

ts/index.ts
wemeetagain: also remove the peer's entry in `_removePeer`

ts/score/peer-score.ts
```diff
@@ -30,6 +30,14 @@ export class PeerScore {
...
    */
   peerIPs: Map<string, Set<string>>
+  /**
+   * Cache score up to decayInterval if topic stats are unchanged.
+   */
+  scoreCache: Map<string, number>
+  /**
+   * The time after which the cached score for a peer is no longer valid.
+   */
+  scoreCacheUntil: Map<string, number>
```
dapplion, line 40: Clarify exactly what number is: `unix timestamp in miliseconds`

dapplion: Consider merging `scoreCache` and `scoreCacheUntil`, it will reduce the memory cost and require a single set and prune operation

ts/index.ts
```diff
@@ -449,7 +467,33 @@ class Gossipsub extends Pubsub {
...
+      return true
+    }
+
+    const score = this.score.score(id)
+    if (score >= ACCEPT_FROM_WHITE_LIST_THRESHOLD_SCORE) {
+      // peer is unlikely to be able to drop its score to `graylistThreshold`
+      // after 128 messages or 1s
+      this.acceptFromWhitelist.set(id, {
+        messagesAccepted: 0,
+        acceptUntil: now + ACCEPT_FROM_WHITE_LIST_DURATION_MS
```
dapplion, line 490: Really nice solution :+1: :fire:

ts/index.ts
tuyennhv: @wemeetagain  so I removed it in `_removePeer` but regarding this

> I think we need a place where these entries get routinely pruned so they don't build up.
Maybe you can prune old entries on every N heartbeats

I don't think it's necessary as we only do the prune in `_removePeer()` function for other map as well (unless there's already a leak somewhere, we'll have to do the prune regularly for all of the map). In `acceptFrom()` we also do the prune if the score is negative

test/accept-from.spec.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+    expect(scoreSpy.getCall(1)).to.be.undefined
+
+    // max messages reached
+    gossipsub._acceptFrom("peerA")
+    expect(scoreSpy.getCall(1).args[0]).to.be.equal("peerA")
+    expect(scoreSpy.getCall(2)).to.be.undefined
+  })
+
+  // TODO: run this in a unit test setup
```
tuyennhv, line 73: there's an issue that sinon causes the test to never finish in `firefox` environment. I think current test set up is like an e2e test where we'll run it in different OSs and browsers, and sinon is not expected to run on it, we need a separate `test:unit` where we can stub any functions

wemeetagain (approved):
LGTM

dapplion (approved):
Looks great!

PR #179, opened 2021-12-22
Fast messsage id cache

tuyennhv:
**Motivation**
+ Gossipsub is designed to have message redundancy, right now we have to calculate 3.97 times per message data in average, in the worse case we calculate 10/11 times per message data

**Description**
+ Follow other clients to have a cheaper faster message id just for message de-duplication, so we have fast message id and canonical message id
+ Key as the sha256 of the message (fast message id), value as the canonical message id => enhance `SimpleTimeCache` for this
+ Instead of passing `getMsgId()` function to other classes (PeerScore, GossipTracer, MessageCache) in the constructor, just pass message id through method as we always have them in the consumer code which make it simpler (and Lighthouse does this!)
+ Use the new `getCanonicalMessageIdStr()` function in all places instead of `getMsgId()` since we can't base on `getMsgId()` in lodestar, the message is removed very quickly from WeakMap
+ `getCanonicalMessageIdStr()`: get from the newly created `getCachedMsgId()` first, then the fast message id cache, then `getMsgId`

**Test result**
+ Message id is calculated 1.02 times per message, max is 2
+ This uses the default cache time which is 30s, should we increase it? Lighthouse uses a duration of 33 slots https://github.com/sigp/lighthouse/blob/fff01b24ddedcd54486e374460855ca20d3dd232/beacon_node/eth2_libp2p/src/config.rs#L36 . We can change the `seenTTL` option anyway
+ With this 30s cache, it takes 1.2MB (and 9556 cache entries) if I use `subscribeAllSubnets` flag in lodestar
Screen Shot 2021-12-22 at 11 01 13

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
   "homepage": "https://github.com/ChainSafe/js-libp2p-gossipsub#readme",
   "dependencies": {
     "@types/debug": "^4.1.7",
-    "@chainsafe/as-sha256": "^0.2.4",
     "debug": "^4.3.1",
     "denque": "^1.5.0",
     "err-code": "^3.0.1",
@@ -53,6 +52,7 @@
   },
   "devDependencies": {
     "@chainsafe/libp2p-noise": "^4.1.1",
+    "@chainsafe/as-sha256": "^0.2.4",
     "@types/chai": "^4.2.3",
     "@types/mocha": "^8.2.2",
     "@typescript-eslint/eslint-plugin": "^3.0.2",
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   "homepage": "https://github.com/ChainSafe/js-libp2p-gossipsub#readme",
   "dependencies": {
     "@types/debug": "^4.1.7",
+    "@chainsafe/as-sha256": "^0.2.4",
     "debug": "^4.3.1",
     "denque": "^1.5.0",
     "err-code": "^3.0.1",
--- a/test/compliance.spec.js
+++ b/test/compliance.spec.js
@@ -6,7 +6,8 @@ const tests = require('libp2p-interfaces-compliance-tests/src/pubsub')
 const Gossipsub = require('../src')
 const { createPeers } = require('./utils/create-peer')
 
-describe('interface compliance', () => {
+describe('interface compliance', function () {
+  this.timeout(3000)
   let peers
   let pubsubNodes = []
 
@@ -17,6 +18,9 @@ describe('interface compliance', () => {
       peers.forEach((peer) => {
         const gossipsub = new Gossipsub(peer, {
           emitSelf: true,
+          // we don't want to cache anything, spec test sends duplicate messages and expect
+          // peer to receive all.
+          seenTTL: -1,
           ...options
         })
 
--- a/test/compliance.spec.js
+++ b/test/compliance.spec.js
@@ -5,6 +5,7 @@ const tests = require('libp2p-interfaces-compliance-tests/src/pubsub')
 
 const Gossipsub = require('../src')
 const { createPeers } = require('./utils/create-peer')
+const { fastMsgIdFn } = require('./utils/msgId')
 
 describe('interface compliance', function () {
   this.timeout(3000)
@@ -18,6 +19,7 @@ describe('interface compliance', function () {
       peers.forEach((peer) => {
         const gossipsub = new Gossipsub(peer, {
           emitSelf: true,
+          fastMsgIdFn,
           // we don't want to cache anything, spec test sends duplicate messages and expect
           // peer to receive all.
           seenTTL: -1,
--- a/test/compliance.spec.js
+++ b/test/compliance.spec.js
@@ -5,7 +5,6 @@ const tests = require('libp2p-interfaces-compliance-tests/src/pubsub')
 
 const Gossipsub = require('../src')
 const { createPeers } = require('./utils/create-peer')
-const { fastMsgIdFn } = require('./utils/msgId')
 
 describe('interface compliance', function () {
   this.timeout(3000)
@@ -19,7 +18,6 @@ describe('interface compliance', function () {
       peers.forEach((peer) => {
         const gossipsub = new Gossipsub(peer, {
           emitSelf: true,
-          fastMsgIdFn,
           // we don't want to cache anything, spec test sends duplicate messages and expect
           // peer to receive all.
           seenTTL: -1,
--- a/test/floodsub.spec.js
+++ b/test/floodsub.spec.js
@@ -17,8 +17,7 @@ const {
   expectSet,
   first,
   startNode,
-  stopNode,
-  fastMsgIdFn
+  stopNode
 } = require('./utils')
 
 describe('gossipsub fallbacks to floodsub', () => {
@@ -27,7 +26,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     let nodeFs
 
     beforeEach(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
       nodeFs = await createFloodsubNode(await createPeer({ peerId: await PeerId.create(), started: false }))
 
       await Promise.all([
@@ -57,7 +56,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     let nodeFs
 
     before(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: false, fastMsgIdFn })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: false })
       nodeFs = await createFloodsubNode(await createPeer({ peerId: await PeerId.create(), started: false }))
 
       await Promise.all([
@@ -90,7 +89,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     let nodeFs
 
     before(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
       nodeFs = await createFloodsubNode(await createPeer({ peerId: await PeerId.create(), started: false }))
 
       await Promise.all([
@@ -141,7 +140,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     const topic = 'Z'
 
     beforeEach(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
       nodeFs = await createFloodsubNode(await createPeer({ peerId: await PeerId.create(), started: false }))
 
       await Promise.all([
@@ -204,7 +203,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     const topic = 'Z'
 
     beforeEach(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
       nodeFs = await createFloodsubNode(await createPeer({ peerId: await PeerId.create(), started: false }))
 
       await Promise.all([
--- a/test/floodsub.spec.js
+++ b/test/floodsub.spec.js
@@ -17,7 +17,8 @@ const {
   expectSet,
   first,
   startNode,
-  stopNode
+  stopNode,
+  fastMsgIdFn
 } = require('./utils')
 
 describe('gossipsub fallbacks to floodsub', () => {
@@ -26,7 +27,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     let nodeFs
 
     beforeEach(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
       nodeFs = await createFloodsubNode(await createPeer({ peerId: await PeerId.create(), started: false }))
 
       await Promise.all([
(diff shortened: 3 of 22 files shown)


ts/index.ts
```diff
@@ -1073,6 +1088,30 @@ class Gossipsub extends Pubsub {
...
+    return this.fastMsgIdCache.get(fastMsgIdStr) ?? messageIdToString(await this.getMsgId(message))
+  }
+
+  /**
+   * Application should override this to return its cached message id without computing it.
+   * Return undefine if it does not have it
+   * @param {InMessage} message
+   * @returns
+   */
+  getCachedMsgId (message: InMessage): Uint8Array | undefined {
```
tuyennhv, line 1111: Lodestar  `getMsgId()` should divide its implementation into `getCachedMsgId()` and `computeMsgId()`

test/peer-score.spec.js
```diff
@@ -520,13 +521,13 @@ describe('PeerScore', () => {
...
-    await ps.validateMessage(msg)
+    await ps.validateMessage(messageIdToString(getMsgId(msg)))
 
     // and reject the message after a duplicate has arrived
     msg.receivedFrom = peerB
-    await ps.duplicateMessage(msg)
+    await ps.duplicateMessage(msg, messageIdToString(getMsgId(msg)))
     msg.receivedFrom = peerA
-    await ps.rejectMessage(msg, ERR_TOPIC_VALIDATOR_REJECT)
+    await ps.rejectMessage(msg, messageIdToString(getMsgId(msg)), ERR_TOPIC_VALIDATOR_REJECT)
```
dapplion, line 530: Define a test utility `getMsgIdStr` to prevent duplication

dapplion (dismissed):
This looks really nice! Thanks for looking into the memory costs :+1: 

Can you please open an issue to raise the timeout to 33 slots? Memory should be analyzed for that case too

ts/index.ts
```diff
@@ -446,17 +453,21 @@ class Gossipsub extends Pubsub {
    * @returns {Promise<void>}
    */
   async _processRpcMessage (msg: InMessage): Promise<void> {
-    const msgID = await this.getMsgId(msg)
-    const msgIdStr = messageIdToString(msgID)
+    const fastMsgIdStr = messageIdToString(SHA256.digest(msg.data))
```
wemeetagain, line 456: I think the fast message id function should be pulled into a method so it may be overridden.

wemeetagain (changes requested):

package.json
```diff
@@ -41,6 +41,7 @@
   "homepage": "https://github.com/ChainSafe/js-libp2p-gossipsub#readme",
   "dependencies": {
     "@types/debug": "^4.1.7",
+    "@chainsafe/as-sha256": "^0.2.4",
```
wemeetagain, line 44: Maybe the fastMsgId should be passed into the constructor, like the msgId in the pubsub base class, so we don't need to add this dependency (maybe just dev dependency for testing).

dapplion (commented):
Looks good! Agree with Cayman suggestion to pull sha256 dependency

dapplion (dismissed):

ts/index.ts
```diff
@@ -146,7 +152,7 @@ class Gossipsub extends Pubsub {
    */
   constructor (
     libp2p: Libp2p,
-    options: Partial<GossipInputOptions> = {}
+    options: Partial<GossipInputOptions> & Pick<GossipInputOptions, 'fastMsgIdFn'>
```
wemeetagain, line 155: We should allow undefined `fastMsgIdFn`

And default to `(msg) => messageIdToString(options.msgIdFn(msg))`

tuyennhv: @wemeetagain `options. msgIdFn` is also optional and I don't see it being used, is that an issue? I suppose we should:
+ remove `msgIdFn` in `GossipInputOptions`
+ default the fast msg id fn to `(msg) => messageIdToString(await this.getMsgId(msg))`

dapplion (approved):

wemeetagain:
I think the fast id feature should be optional. Likely many applications (including ipfs and existing apps) will not need this feature and shouldn't pay for the cost.
We can switch some of this logic based on if the fast id function is set or not. eg: `if (this.fastIdFnStr) { ... } else { ... }`

dapplion (approved):

test/heartbeat.spec.js
```diff
@@ -8,14 +8,15 @@ const { GossipsubHeartbeatInterval } = require('../src/constants')
 const {
   createPeer,
   startNode,
-  stopNode
+  stopNode,
+  fastMsgIdFn,
```
wemeetagain, line 12:
```suggestion
  stopNode
```

test/heartbeat.spec.js
```diff
@@ -8,14 +8,15 @@ const { GossipsubHeartbeatInterval } = require('../src/constants')
...
 } = require('./utils')
 
 describe('heartbeat', () => {
   let gossipsub
 
   before(async () => {
-    gossipsub = new Gossipsub(await createPeer({ started: false }), { emitSelf: true })
+    gossipsub = new Gossipsub(await createPeer({ started: false }), { emitSelf: true, fastMsgIdFn })
```
wemeetagain, line 19:
```suggestion
    gossipsub = new Gossipsub(await createPeer({ started: false }), { emitSelf: true })
```

test/compliance.spec.js
```diff
@@ -5,8 +5,10 @@ const tests = require('libp2p-interfaces-compliance-tests/src/pubsub')
 
 const Gossipsub = require('../src')
 const { createPeers } = require('./utils/create-peer')
+const { fastMsgIdFn } = require('./utils/msgId')
```
wemeetagain, line 8:
```suggestion
```

test/compliance.spec.js
```diff
@@ -17,6 +19,10 @@ describe('interface compliance', () => {
       peers.forEach((peer) => {
         const gossipsub = new Gossipsub(peer, {
           emitSelf: true,
+          fastMsgIdFn,
```
wemeetagain, line 22:
```suggestion
```

test/floodsub.spec.js
```diff
@@ -17,7 +17,8 @@ const {
   expectSet,
   first,
   startNode,
-  stopNode
+  stopNode,
+  fastMsgIdFn
```
wemeetagain, line 21:
```suggestion
  stopNode
```

test/floodsub.spec.js
```diff
@@ -26,7 +27,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     let nodeFs
 
     beforeEach(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
```
wemeetagain, line 30:
```suggestion
      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
```

test/floodsub.spec.js
```diff
@@ -56,7 +57,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     let nodeFs
 
     before(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: false })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: false, fastMsgIdFn })
```
wemeetagain, line 60:
```suggestion
      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: false })
```

test/floodsub.spec.js
```diff
@@ -89,7 +90,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     let nodeFs
 
     before(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
```
wemeetagain, line 93:
```suggestion
      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
```

test/floodsub.spec.js
```diff
@@ -140,7 +141,7 @@ describe('gossipsub fallbacks to floodsub', () => {
     const topic = 'Z'
 
     beforeEach(async () => {
-      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
+      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true, fastMsgIdFn })
```
wemeetagain, line 144:
```suggestion
      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
```

test/floodsub.spec.js
```diff
@@ -203,7 +204,7 @@ describe('gossipsub fallbacks to floodsub', () => {
...
```
wemeetagain, line 207:
```suggestion
      nodeGs = new Gossipsub(await createPeer({ started: false }), { fallbackToFloodsub: true })
```

ts/index.ts
```diff
@@ -146,7 +151,7 @@ class Gossipsub extends Pubsub {
    */
   constructor (
     libp2p: Libp2p,
-    options: Partial<GossipInputOptions> = {}
+    options: Partial<GossipInputOptions>
```
wemeetagain, line 154:
```suggestion
    options: Partial<GossipInputOptions> = {}
```

ts/index.ts
```diff
@@ -204,11 +209,12 @@ class Gossipsub extends Pubsub {
     })
 
     /**
-     * Cache of seen messages
+     * The canonical cache of seen messages, this goes with getMsgId() function.
+     * Only compute getMsgId() once per message even with the redundantcy.
```
wemeetagain, line 213:
```suggestion
     * Cache of seen messages
```

ts/index.ts
```diff
@@ -277,9 +283,18 @@ class Gossipsub extends Pubsub {
 
     /**
      * A message cache that contains the messages for last few hearbeat ticks
-     *
      */
-    this.messageCache = options.messageCache || new MessageCache(opts.mcacheGossip, opts.mcacheLength, this.getMsgId.bind(this))
+    this.messageCache = options.messageCache || new MessageCache(opts.mcacheGossip, opts.mcacheLength)
+
+    /**
+     * A fast message id function which return a string from InMessage
```
wemeetagain, line 290:
```suggestion
     * A fast message id function used for internal message de-duplication
```

ts/index.ts
```diff
@@ -277,9 +283,18 @@ class Gossipsub extends Pubsub {
...
+     */
+    this.getFastMsgIdStr = options.fastMsgIdFn ?? undefined
+
+    /**
+     * This uses a faster cheaper message-id function for internal message de-duplication.
```
wemeetagain, line 295:
```suggestion
     * Maps fast message-id to canonical message-id
```

ts/index.ts
```diff
@@ -446,17 +461,33 @@ class Gossipsub extends Pubsub {
...
+      }
+      canonicalMsgIdStr = messageIdToString(await this.getMsgId(msg))
 
-    // Ignore if we've already seen the message
-    if (this.seenCache.has(msgIdStr)) {
-      this.score.duplicateMessage(msg)
-      return
+      // put to cache
+      this.seenCache.put(canonicalMsgIdStr)
+      this.fastMsgIdCache.put(fastMsgIdStr, canonicalMsgIdStr)
```
wemeetagain, line 477:
```suggestion
      this.fastMsgIdCache.put(fastMsgIdStr, canonicalMsgIdStr)
```

ts/index.ts
```diff
@@ -446,17 +461,33 @@ class Gossipsub extends Pubsub {
...
+    } else {
+      // check duplicate
+      canonicalMsgIdStr = messageIdToString(await this.getMsgId(msg))
+      if (this.seenCache.has(canonicalMsgIdStr)) {
+        this.score.duplicateMessage(msg, canonicalMsgIdStr)
+        return
+      }
+
+      // put to cache
+      this.seenCache.put(canonicalMsgIdStr)
```
wemeetagain, line 487:
```suggestion
```

ts/index.ts
```diff
@@ -446,17 +461,33 @@ class Gossipsub extends Pubsub {
...
     }
-    this.seenCache.put(msgIdStr)
 
-    await this.score.validateMessage(msg)
+    await this.score.validateMessage(canonicalMsgIdStr)
```
wemeetagain, line 490:
```suggestion
    // put in cache
    this.seenCache.put(canonicalMsgIdStr)

    await this.score.validateMessage(canonicalMsgIdStr)
```

ts/index.ts
```diff
@@ -1073,6 +1108,29 @@ class Gossipsub extends Pubsub {
...
   }
 
+  /**
+   * Fast message id logic: Get from the application cache first, then from the fast cache, then from this.getMsgId()
+   * Non fast message id: just get from message id
+   * @param {InMessage} msg
+   */
+  async getCanonicalMsgIdStr (msg: InMessage): Promise<string> {
+    return (this.fastMsgIdCache && this.getFastMsgIdStr)
+      ? this.getCachedMsgIdStr(msg) ?? this.fastMsgIdCache.get(await this.getFastMsgIdStr(msg)) ?? messageIdToString(await this.getMsgId(msg))
```
wemeetagain, line 1118:
```suggestion
      ? this.getCachedMsgIdStr(msg) ?? this.fastMsgIdCache.get(this.getFastMsgIdStr(msg)) ?? messageIdToString(await this.getMsgId(msg))
```

ts/index.ts
```diff
@@ -1073,6 +1108,29 @@ class Gossipsub extends Pubsub {
     }
   }
 
+  /**
+   * Fast message id logic: Get from the application cache first, then from the fast cache, then from this.getMsgId()
+   * Non fast message id: just get from message id
+   * @param {InMessage} msg
```
wemeetagain, line 1114:
```suggestion
   * Return the canonical message-id of a message as a string
   *
   * If a fast message-id is set: Try 1. the application cache 2. the fast cache 3. `getMsgId()`
   * If a fast message-id is NOT set: Just `getMsgId()`
   * @param {InMessage} msg
   * @returns {Promise<string>}
```

ts/index.ts
```diff
@@ -1073,6 +1108,29 @@ class Gossipsub extends Pubsub {
...
+      : messageIdToString(await this.getMsgId(msg))
+  }
+
+  /**
+   * Fast message id logic:
+   *   Application should override this to return its cached message id string without computing it.
+   *   Return undefined if it does not have it
+   * Non fast message id logic: application doesn't need to override this
+   * @param {InMessage} msg
+   * @returns
```
wemeetagain, line 1128:
```suggestion
   * An application should override this function to return its cached message id string without computing it.
   * Return undefined if message id is not found.
   * If a fast message id function is not defined, this function is ignored.
   * @param {InMessage} msg
   * @returns {string | undefined}
```

wemeetagain (changes requested):

wemeetagain (approved):

PR #182, opened 2021-12-31
feat: async peerstore

achingbrain:
Refactors interfaces and classes used by `libp2p-interfaces` to use the async peer store from https://github.com/libp2p/js-libp2p/pull/1058

Fixes a memory leak where peer data (multiaddrs, protocols, etc) is never evicted from memory.

BREAKING CHANGE: peerstore methods and pubsub start/stop are now all async

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "denque": "^1.5.0",
     "err-code": "^3.0.1",
     "it-pipe": "^1.1.0",
-    "libp2p-interfaces": "^2.0.1",
+    "libp2p-interfaces": "^4.0.4",
     "peer-id": "^0.16.0",
     "protobufjs": "^6.11.2",
     "uint8arrays": "^3.0.0"
@@ -53,6 +53,7 @@
   "devDependencies": {
     "@chainsafe/libp2p-noise": "^4.1.1",
     "@chainsafe/as-sha256": "^0.2.4",
+    "@dapplion/benchmark": "^0.1.6",
     "@types/chai": "^4.2.3",
     "@types/mocha": "^8.2.2",
     "@typescript-eslint/eslint-plugin": "^3.0.2",
@@ -71,9 +72,9 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
     "it-pair": "^1.0.0",
-    "libp2p": "^0.35.0",
-    "libp2p-floodsub": "^0.28.0",
-    "libp2p-interfaces-compliance-tests": "^2.0.3",
+    "libp2p": "libp2p/js-libp2p#feat/async-peerstore",
+    "libp2p-floodsub": "^0.29.0",
+    "libp2p-interfaces-compliance-tests": "^4.0.6",
     "libp2p-mplex": "^0.10.3",
     "libp2p-websockets": "^0.16.1",
     "lodash": "^4.17.15",
@@ -85,7 +86,6 @@
     "promisify-es6": "^1.0.3",
     "sinon": "^11.1.1",
     "time-cache": "^0.3.0",
-    "@dapplion/benchmark": "^0.1.6",
     "typescript": "4.0.x",
     "util": "^0.12.3"
   },
--- a/test/go-gossipsub.js
+++ b/test/go-gossipsub.js
@@ -75,7 +75,7 @@ const awaitEvents = (emitter, event, number, timeout = 10000) => {
   })
 }
 
-describe("go-libp2p-pubsub gossipsub tests", function () {
+describe.skip("go-libp2p-pubsub gossipsub tests", function () {
   this.timeout(100000)
   afterEach(() => {
     sinon.restore()
--- a/test/go-gossipsub.js
+++ b/test/go-gossipsub.js
@@ -2,7 +2,9 @@
  * These tests were translated from:
  *   https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub_test.go
  */
-const { expect } = require('chai')
+const chai = require('chai')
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -70,9 +70,9 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
     "it-pair": "^1.0.0",
-    "libp2p": "^0.35.0",
-    "libp2p-floodsub": "^0.28.0",
-    "libp2p-interfaces-compliance-tests": "^2.0.3",
+    "libp2p": "libp2p/js-libp2p#feat/async-peerstore",
```
achingbrain, line 73: This will need to be removed after the next libp2p release.

Longer term we need to break the circular dependencies of these components.

dapplion:
> Fixes a memory leak where peer data (multiaddrs, protocols, etc) is never evicted from memory.

Where was the leak exactly?

achingbrain:
> Where was the leak exactly?

See the discussion in https://github.com/libp2p/js-libp2p/pull/1058

In brief the peer store has various Maps that we add peer data to when we discover them on the network - multiaddrs, protocols, etc.  The leak is that we never evict anything from the Map so it just grows and grows.

The fix is to use the datastore to store and read peer data instead of holding it all in memory.

achingbrain:
There seem to be quite a few flaky tests here, from what it looks like, the errors are the same as in master & other PRs - @dapplion any ideas?

dapplion:
> The fix is to use the datastore to store and read peer data instead of holding it all in memory.

So everytime libp2p needs to lookup protocols for a peer it will do a database disk read?

achingbrain:
It’ll do a datastore read - that datastore could have an in-memory cache which would replicate the previous behaviour but you could add an eviction strategy to stop your nodes eventually falling over.

dapplion:
I feel a better solution would be to fix the in-memory cache first instead of moving to a disk database. If those values have to be persisted, the app layer can choose when to do so, for example right before shutting down.

achingbrain:
This PR enables the fix for the in-memory cache and does not move to a disk database.  It allows libp2p to use a [datastore](https://github.com/ipfs/js-ipfs-interfaces/tree/master/packages/interface-datastore#readme) which makes the actual storage configurable - it could be [in-memory](https://github.com/ipfs/js-datastore-core/blob/master/src/memory.js) or a [database](https://github.com/ipfs/js-datastore-level) or a [combination of both](https://github.com/ipfs/js-datastore-core/blob/master/src/mount.js).

At the moment peer store access is synchronous, which means you are stuck with in-memory storage since every storage method that's not in-memory requires async accessors.

You then can't evict peer information as there's no way to retrieve it later from a secondary async storage medium without requiring the calling code to null-guard on every peer store access and go to the secondary storage medium instead.  This would add significant complexity to any part of the stack that touches the peer store which is quite a lot of it.

The change here is just to move to async accessors which then lets you configure the datastore as your application requires.

For the next IPFS release this will be an in-memory datastore with a configurable LRU cache that falls back to something that can store larger amounts of data without impacting memory use such as level or IndexedDB while also periodically writing cache data into the store in the same way the current persistent peer store works.

If you want to be able to hold the peer info of every node in the network in memory as we do now, just configure the LRU cache to be something very large.  Even then you'll still gain some protection against OOM errors as the LRU cache will have a limit.

dapplion:
That sounds good @achingbrain ! Thanks for the thorough explanation :+1:

package.json
vasco-santos: We can, this was already done for pubsub as part of https://github.com/libp2p/js-libp2p/issues/857 but we need to do for the other modules. Just opened a PR to remove the dev deps: https://github.com/libp2p/js-libp2p/pull/1107

test/go-gossipsub.js
```diff
@@ -2,7 +2,9 @@
  * These tests were translated from:
  *   https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub_test.go
  */
-const { expect } = require('chai')
+const chai = require('chai')
+chai.use(require('dirty-chai'))
```
vasco-santos, line 6: nit: use aegir exported chai instead

vasco-santos (dismissed):
LGTM

test/go-gossipsub.js
achingbrain: That would be my preference too, but it's not used anywhere else in this module so I did the manual `.use` to be consistent with the other tests.

wemeetagain:
Generally LGTM but there is a type error:
https://github.com/ChainSafe/js-libp2p-gossipsub/runs/4717165393?check_suite_focus=true#step:4:29

achingbrain:
@wemeetagain the type error has been resolved, the CI problems are back to the flakiness observed earlier.

dapplion (dismissed):

wemeetagain:
This PR introduces some subtle difference in timing that results in the tests inconsistently failing.
IMO we need to diagnose the issue/resolve it before merging this

ts/index.ts
```diff
@@ -1467,17 +1467,17 @@ class Gossipsub extends Pubsub {
       const peers = getGossipPeers(this, topic, constants.GossipsubPrunePeers, (xid: string): boolean => {
         return xid !== id && this.score.score(xid) >= 0
       })
-      peers.forEach(p => {
+      for (const p of peers) {
```
wemeetagain, line 1470: This path should use a `Promise.all` to allow multiple `getRawEnvelope` calls to be triggered concurrently.

wemeetagain:
Maybe we can just disable the unreliable tests for now

wemeetagain (approved):

PR #187, opened 2022-02-15
Add getMsgIdIfNotSeen() method

tuyennhv:
**Motivation**
It's important to track how many messages were seen in lodestar

**Description**
+ Refactor `processRpcMessage` to a separate `getMsgIdIfNotSeen()` method so that lodestar can override
+ No functionality changed

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -467,7 +467,7 @@ class Gossipsub extends Pubsub {
    * @returns {Promise<void>}
    */
   async _processRpcMessage(msg: InMessage): Promise<void> {
-    const canonicalMsgIdStr = await this.isSeen(msg)
+    const canonicalMsgIdStr = await this.getMsgIdIfNotSeen(msg)
     // if not seen
     if (canonicalMsgIdStr) {
       // put in cache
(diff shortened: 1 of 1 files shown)


ts/index.ts
```diff
@@ -460,14 +460,28 @@ class Gossipsub extends Pubsub {
...
+      // put in cache
+      this.seenCache.put(canonicalMsgIdStr)
+
+      await this.score.validateMessage(canonicalMsgIdStr)
+      await super._processRpcMessage(msg)
+    }
+  }
+
+  /** Return undefined if we've seen the message, fastMsgIdStr if not seen */
+  async isSeen (msg: InMessage): Promise<string | undefined> {
```
dapplion, line 475: The name is very confusing since responding to the question "is seen?" you expect truthy to mean yes, falsy to mean no.  What about a longer but more correct name like `getMsgIdIfNotSeen`.

dapplion (approved):

tuyennhv:
new gossipsub contains a lot of metrics, no need this anymore

PR #189, opened 2022-02-22
_publish: Only compute score if peer stream has FloodsubID protocol

### LINKED ISSUE #188: Redundant peer score computation when publishing message
**Description**
When publishing message:
```ts
 // floodsub peers
        peersInTopic.forEach((id) => {
          const score = this.score.score(id)
          const peerStreams = this.peers.get(id)
          if (!peerStreams) {
            return
          }
          if (peerStreams.protocol === constants.FloodsubID && score >= this._options.scoreThresholds.publishThreshold) {
            tosend.add(id)
          }
        })
```

the debug shows that we rarely have peer with FloodsubID protocol while we already compute score for peer.

This was taken with gossipsub 0.11.1 and it took >6.5% of lodestar cpu time
[0222_contabo_13_ssz_v2.cpuprofile.zip](https://github.com/ChainSafe/js-libp2p-gossipsub/files/8114067/0222_contabo_13_ssz_v2.cpuprofile.zip)

**Solution**
Only compute peer score when peer has FloodsubID protocol

tuyennhv:
**Motivation**
+ Avoid computing score when it's not really necessary

**Description:**
+ _publish: Only compute score if peer stream has FloodsubID protocol
+ It's not a big improvement with `master` branch since we cache score already
+ There's another `this.score.score(id)` call some lines below but it's ok since we have score cache
Close #188

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -1176,12 +1176,11 @@ class Gossipsub extends Pubsub {
 
         // floodsub peers
         peersInTopic.forEach((id) => {
-          const score = this.score.score(id)
           const peerStreams = this.peers.get(id)
           if (!peerStreams) {
             return
           }
-          if (peerStreams.protocol === constants.FloodsubID && score >= this._options.scoreThresholds.publishThreshold) {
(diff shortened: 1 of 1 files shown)


dapplion (approved):
LGTM

wemeetagain (approved):

PR #200, opened 2022-03-11
chore: increase the default for seenTTL to match that of go-libp2p

stbrody:
In the go implementation of gossipsub, libp2p keeps a 2 minute long cache of recently seen pubsub messages so that the node won't forward on duplicate messages.  In the JS implementation the default seen cache ttl is only 30 seconds.  We have seen issues with duplicate pubsub messages (see https://github.com/libp2p/js-libp2p/issues/1043) and think this shorter cache ttl may be part of the problem.  In any case it would be good for the default behavior to be consistent between the javascript and go implementation


wemeetagain (approved):
LGTM

PR #212, opened 2022-03-20
Do not apply P3 penalty if peers are pruned from topic mesh

tuyennhv:
**Motivation**
+ Part of the effort to investigate why P3 is applied a lot to peers

**Description**
+ Do not apply P3 penalty if peers are pruned from topic mesh (double checked with rust-libp2p)

### DIFF
--- a/ts/score/peer-score.ts
+++ b/ts/score/peer-score.ts
@@ -237,6 +237,7 @@ export class PeerScore {
       }
 
       tstats.inMesh = false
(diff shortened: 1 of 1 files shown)


dapplion (approved):

tuyennhv:
Right now P3 is applied heavily even after peers left mesh (there's only 1 mesh peer)

Screen Shot 2022-03-21 at 10 49 47

PR #214, opened 2022-03-23
Forward messages to floodsub peers

tuyennhv:
**Motivation**
+ Spec:
```
If the message is valid, was not published by the router itself, and has not been previously seen, the router will forward the message. First, it will forward the message to every peer in peers.floodsub[topic] for backwards-compatibility with floodsub. Next, it will forward the message to every peer in its local gossipsub topic mesh, contained in mesh[topic].
```

+ Fix "test mixed gossipsub" in go-gossipsub test

**Description**
+ Maintain floodsub peers in a set
+ Handle it in "onConnect", "onDisconnect"
+ Use it when selecting peers to forward/publish

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -1627,7 +1627,9 @@ export default class Gossipsub extends EventEmitter {
         }
       })
 
-      // as of Mar 2022, rust-libp2p does not have this while golang-libp2p and the spec do have
+      // As of Mar 2022, spec + golang-libp2p include this while rust-libp2p does not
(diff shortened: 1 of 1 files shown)


ts/index.ts
```diff
@@ -1673,14 +1694,8 @@ export default class Gossipsub extends EventEmitter {
...
 
         // floodsub peers
-        peersInTopic.forEach((id) => {
-          const peerStreams = this.peers.get(id)
-
-          if (
-            peerStreams &&
-            peerStreams.protocol === constants.FloodsubID &&
-            this.score.score(id) >= this.opts.scoreThresholds.publishThreshold
-          ) {
```
tuyennhv, line 1683: Some notes
+ If there is no floodsub peers, we save a loop through topic peers
+ One difference is the check of `peersInTopic.has(id)`, master has this but rust-libp2p does not have

dapplion (approved):

Issue #235, opened 2022-04-26
Rebalance mesh on receipt of peer subscriptions?


achingbrain:
When the subscriptions of a peer are sent, we add them to the subscribers for a given topic, but we have to wait for the heartbeat before they (might) get added to the mesh for the topic, so there's a brief window during which we know a peer is interested in a topic, but we won't gossip any messages on that topic to them.

I'm not sure why it's like this, we could rebalance the mesh when we are informed of their subscriptions and before emitting the `subscription-change` event, then [arbitrary delays in tests like this](https://github.com/libp2p/js-libp2p-interfaces/pull/202) would not be necessary which makes everything a lot less flaky.

dapplion:
@achingbrain That's a good point! Shouldn't be expensive to do

Issue #183, opened 2022-01-20
Re-enable unreliable tests


wemeetagain:
in #182 we disabled tests in `test/go-gossipsub.js` because several tests would fail in CI

We should figure our what's failing in the tests and re-enable them

dapplion:
This seems very urgent, it's worrying that some many tests break just by changing the API to using promises

wemeetagain:
Well this is the interesting part: they don't break locally, and they don't _always_ break in CI. And they don't always break in the same place or in the same tests.

So it seems more that the CI environment sucks and the tests are unreliable and relying on specific timing (that is the case in normal operation, but not so in sucky CI).

This is evident when the heartbeat test failed, like [here](https://github.com/ChainSafe/js-libp2p-gossipsub/runs/4674180733?check_suite_focus=true#step:6:134). A timer set for 1000ms took 1124ms to return.

dapplion:
The danger is that a new feature PR might break something and this tests wont sound the alarm, allowing bugs to releases.

Extending the timeouts to x5 could help?

PR #236, opened 2022-04-26
fix: update to esm libp2p and deps

achingbrain:
- Updates all deps to `@libp2p/*` versions
- Updates config to output ESM
- Makes the `GossipSub` class implement the `PubSub` interface so it can be used with `libp2p@next`
- Removes the `libp2p` constructor arg as the components are injected by `libp2p` at runtime
- Swaps protobufjs for protons

BREAKING CHANGE: the output of this module is now ESM-only


achingbrain:
This is not 100% ready for merge just yet, but I wanted to get some early feedback.

Some thoughts:

- I had to make a load of private methods public in the `GossipSub` class for the tests to compile.  I figure this isn't the end of the world since at runtime `libp2p` uses the `PubSub` interface for the public API and not `GossipSub`
- Because `libp2p` is no longer responsible for the lifecycle of it's components (e.g. the user news them up instead of passing constructors & constructor args in which are essentially impossible to type), we can't pass a `libp2p` node in to the `GossipSub` constructor any more, instead it now gets the components it needs by being [Initializable](https://github.com/libp2p/js-libp2p-interfaces/blob/ec81622e5b7c6d256e0f8aed6d3695642473293b/packages/libp2p-interfaces/src/components.ts#L16-L18).  The tests used to access the `.libp2p` prop of `GossipSub` but now they use instances of `libp2p` and access `GossipSub` via `libp2p.pubusb` instead so it's been flipped around.  The access of `.libp2p` in the tests is mostly used to connect `GossipSub` instances together - longer term this should be refactored to do the same thing via the Registrar component (see the [pubsub interface compliance tests for an example](https://github.com/libp2p/js-libp2p-interfaces/blob/ec81622e5b7c6d256e0f8aed6d3695642473293b/packages/libp2p-interface-compliance-tests/src/pubsub/two-nodes.ts#L69-L75)) - this is much lighter weight, is faster and means this module doesn't need to depend on `libp2p`, transports, encryption, multiplexers, etc, etc, any more.
- This module uses `aegir` for (I think) pretty much just releases and building a browser bundle now, since it has it's own eslint rules, ts configuration etc, so it's getting a lot of the costs of `aegir` with very little of the benefit.  I think it should either be all-in on `aegir` or shouldn't use it at all.  I am more than happy to add extra linting/ts rules to `aegir` as required (enforcing return types on methods is definitely something we should add, for example).

achingbrain:
Outstanding items of work:

1. Resolve conflicts 😩 
2. Tests passing - the last few are in the `go-libp2p-pubsub gossipsub tests` suite - there seem to be some race conditions in the tests as it's the classic "fail when run together", "pass when run individually" thing.  Changes like that proposed in #235 may help here.  Looks like they've been flaky before #183 😅

wemeetagain:
Once ESM-only lands, we can rename directories to be consistent with other libp2p repos.

The reason we had source (typescript) files under `ts/`, and build artifacts (js) files under `src/` was to allow consumers to import subdirectories from the "same place" as other libp2p packages. eg: `import foo from "libp2p-gossip/src/foo"` being similar to `import bar from "libp2p/src/bar"`.
After ESM-only, exports must be named explicitly in package.json, and the `src/` subdirectory name no longer bleeds into the export name.

package.json
```diff
@@ -3,26 +3,28 @@
   "version": "0.14.0",
```
wemeetagain, line 3: Do we want to update the package name to use the @libp2p namespace?

achingbrain: Maybe use the @chainsafe namespace same as noise?

dapplion: I really feel that for the sake of consistency having all packages under the @libp2p namespace has way more positives. The ethos of libp2p is composability. So a LEGO piece being in a foreign namespace, makes a dis-service to that goal.

I would much prefer to change `libp2-gossipsub` package name to `@libp2p/gossipsub`

BigLep: Per 2022-05-10 verbal conversation during JS triage: it was agreed to keep using the chainsafe namespace.  @wemeetagain can leave other comments.

mpetrunic (approved):

PR #258, opened 2022-06-02
fix: flaky tests, enable firefox tests in ci

mpetrunic:
It would seem that during long-running tests on weak hardware, peers don't respond in a timely manner and they get penalties and pruned which stalls tests.


wemeetagain:
closing in favor of #259

Issue #256, opened 2022-05-27
heartbeat performance issue


tuyennhv:
**Motivation**

In a contabo node of 50 peers, `heartbeat` function may take up to 800ms - 1s which block the event loop

<img width="1249" alt="Screen Shot 2022-05-27 at 16 59 20" src="https://user-images.githubusercontent.com/10568965/170677668-aef3a8f6-13c0-4825-9933-557f816bc70a.png">

part of https://github.com/ChainSafe/lodestar/issues/4002

**Description**
- Write performance test for `heartbeat()` function
- Improve `heartbeat()` performance
- If it takes too much time, yield to event loop every x milli seconds

dapplion:
@tuyennhv The resolution of the buckets in `gossipsub_heartbeat_duration_seconds_bucket` is really poor: [0.01, 0.1, 1]. So the quantile function can show widely inaccurate results. We should improve the bucket values and check this metric again. Since heartbeats are about 1s and we care about not looping around better bucket values would be
```
[0.01, 0.05, 0.1, 0.2, 0.4, 0.7, 1]
```
As `gossipsub_heartbeat_duration_seconds_bucket` doesn't have any labels and is called infrequently, more buckets are cheap

dapplion:
![Screenshot from 2022-05-28 14-12-48](https://user-images.githubusercontent.com/35266934/170814886-e4310e12-f712-4af3-9b5b-09cddfd79431.png)

This chart is also a bit worrying: heartbeat should not be skipped, and should run 1/0.7 = 1.42 times / sec on average. Smooth chart below

![Screenshot from 2022-05-28 14-14-43](https://user-images.githubusercontent.com/35266934/170814970-ebf07264-3834-4ea7-a8d6-3fac89022fe8.png)

dapplion:
The only hard data we have now is how many heartbeats take > 100ms to complete and how many take > 1000ms to complete. However you can see that many take > 100ms but almost 0 take > 1000 ms. But we don't really now the distribution between 100 and 1000 at all :shrug: 

![Screenshot from 2022-05-28 14-26-19](https://user-images.githubusercontent.com/35266934/170815377-8939ff3a-9cf5-4b7c-97dd-b0c2c9805cdb.png)
![Screenshot from 2022-05-28 14-26-36](https://user-images.githubusercontent.com/35266934/170815378-09ea3083-53fc-485d-a680-56be749f9258.png)

[...]

PR #268, opened 2022-06-07
perf: improve heartbeat performance

tuyennhv:
**Motivation**
+ heartbeat() takes time as shown in #256 

**Description**
- Right now in heartbeat we call `getRandomGossipPeers` multiple times to:
  - get more mesh peers
  - more outbound peers
  - opportunistic grafting
  - emit gossip

we should loop through the topics peers only 1 time, it's an almost 2x performance as shown in the `heartbeat()` benchmark

- part of #256

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2085,12 +2085,22 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
   }
 
   /**
+   * Emits gossip - Send IHAVE messages to a random set of gossip peers. This is applied to mesh
+   * and fanout peers
+   */
+  private emitGossip(peersToGossipByTopic: Map<string, Set<PeerIdStr>>): void {
+    for (const [topic, peersToGossip] of peersToGossipByTopic) {
+      this.doEmitGossip(topic, peersToGossip)
+    }
+  }
+
+  /**
    * Emits gossip to peers in a particular topic
    *
    * @param topic
-   * @param exclude - peers to exclude
+   * @param candidateToGossip - peers to gossip
    */
-  private emitGossip(topic: string, exclude: Set<string>): void {
+  private doEmitGossip(topic: string, candidateToGossip: Set<PeerIdStr>): void {
     const messageIDs = this.mcache.getGossipIDs(topic)
     if (!messageIDs.length) {
       return
@@ -2109,39 +2119,23 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     // First we collect the peers above gossipThreshold that are not in the exclude set
     // and then randomly select from that set
     // We also exclude direct peers, as there is no reason to emit gossip to them
-    const peersToGossip: string[] = []
-    const topicPeers = this.topics.get(topic)
-    if (!topicPeers) {
-      // no topic peers, no gossip
-      return
-    }
-    topicPeers.forEach((id) => {
-      const peerStreams = this.peers.get(id)
-      if (!peerStreams) {
-        return
-      }
-      if (
-        !exclude.has(id) &&
-        !this.direct.has(id) &&
-        hasGossipProtocol(peerStreams.protocol) &&
-        this.score.score(id) >= this.opts.scoreThresholds.gossipThreshold
-      ) {
-        peersToGossip.push(id)
-      }
-    })
 
+    if (!candidateToGossip.size) return
     let target = this.opts.Dlazy
-    const factor = constants.GossipsubGossipFactor * peersToGossip.length
+    const factor = constants.GossipsubGossipFactor * candidateToGossip.size
+    let peersToGossip: Set<PeerIdStr> | PeerIdStr[] = candidateToGossip
     if (factor > target) {
       target = factor
     }
-    if (target > peersToGossip.length) {
-      target = peersToGossip.length
+    if (target > peersToGossip.size) {
+      target = peersToGossip.size
     } else {
-      shuffle(peersToGossip)
+      // only shuffle if needed
+      peersToGossip = shuffle(Array.from(peersToGossip)).slice(0, target)
     }
+
     // Emit the IHAVE gossip to the selected peers up to the target
-    peersToGossip.slice(0, target).forEach((id) => {
+    peersToGossip.forEach((id) => {
       let peerMessageIDs = messageIDs
       if (messageIDs.length > constants.GossipsubMaxIHaveLength) {
         // shuffle and slice message IDs per peer so that we emit a different set for each peer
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -2109,39 +2119,23 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     // First we collect the peers above gossipThreshold that are not in the exclude set
     // and then randomly select from that set
     // We also exclude direct peers, as there is no reason to emit gossip to them
```
wemeetagain, line 2121: move these comments to somewhere thats more relevant or delete if there's nowhere that makes sense

src/index.ts
```diff
@@ -2084,13 +2084,23 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     }
   }
 
+  /**
+   * Emits gossip - Send IHAVE messages to a random set of gossip peers. This is applied to mesh
```
wemeetagain, line 2088: I'm pretty sure we _don't_ gossip to mesh peers, but rather only to non-mesh peers (that meet the other criteria)

wemeetagain, line 2088: I'm pretty sure we _don't_ gossip to mesh peers, but rather only to non-mesh peers (that meet the other criteria)

src/index.ts
```diff
@@ -2310,8 +2304,56 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
...
+
+      const meshPeersFromCandidates = (ineed: number, cond: (peer: PeerIdStr) => boolean): PeerIdStr[] => {
+        let count = 0
+        let index = 0
+        const newMeshPeers = []
+
+        while (count < ineed && index < candidateMeshPeers.length) {
+          if (cond(candidateMeshPeers[index])) {
+            newMeshPeers.push(candidateMeshPeers[index])
+            candidateMeshPeers.splice(index, 1)
```
wemeetagain, line 2347: I think splicing arrays is somewhat slow as it has too reindex all following values.

Maybe try making candidateMeshPeers a Set. It should iterate thru in insertion order (and won't need to reindex as you remove items from it)

newMeshPeers could also be a Set

Worth benchmarking and seeing

wemeetagain, line 2347: I think splicing arrays is somewhat slow as it has too reindex all following values.

Maybe try making candidateMeshPeers a Set. It should iterate thru in insertion order (and won't need to reindex as you remove items from it)

src/index.ts
```diff
@@ -2320,6 +2362,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
         this.addBackoff(id, topic)
         // remove peer from mesh
         peers.delete(id)
+        if (getScore(id) >= this.opts.scoreThresholds.gossipThreshold) peersToGossip.add(id)
```
wemeetagain, line 2365: why do we need this now?

wemeetagain, line 2365: can you add a comment for this?

wemeetagain, line 2365: can you add a comment for this?

src/index.ts
```diff
@@ -2336,6 +2379,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
         this.score.graft(id, topic)
         // add peer to mesh
         peers.add(id)
+        peersToGossip.delete(id)
```
wemeetagain, line 2382: and a comment for this

wemeetagain, line 2382: and a comment for this

src/index.ts
```diff
@@ -2310,8 +2304,56 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
...
+
+      if (peersInTopic) {
+        const shuffledPeers = shuffle(Array.from(peersInTopic))
+        const backoff = this.backoff.get(topic)
+        for (const id of shuffledPeers) {
+          const peerStreams = this.peers.get(id)
+          if (peerStreams && hasGossipProtocol(peerStreams.protocol) && !peers.has(id) && !this.direct.has(id)) {
+            const score = getScore(id)
+            if ((!backoff || !backoff.has(id)) && score >= 0) candidateMeshPeers.push(id)
+            if (score >= this.opts.scoreThresholds.gossipThreshold) peersToGossip.add(id)
```
wemeetagain, line 2334: can you comment whats happening here?

wemeetagain, line 2334: can you comment whats happening here?

src/index.ts
```diff
@@ -2506,25 +2535,36 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
...
+      peersToGossipByTopic.set(topic, peersToGossip)
+
+      if (peersInTopic) {
+        const shuffledPeers = shuffle(Array.from(peersInTopic))
+        for (const id of shuffledPeers) {
+          const peerStreams = this.peers.get(id)
+          if (peerStreams && hasGossipProtocol(peerStreams.protocol) && !fanoutPeers.has(id) && !this.direct.has(id)) {
+            const score = getScore(id)
+            if (score >= this.opts.scoreThresholds.publishThreshold) candidateFanoutPeers.push(id)
+            if (score >= this.opts.scoreThresholds.gossipThreshold) peersToGossip.add(id)
```
wemeetagain, line 2551: comment for this

wemeetagain, line 2551: comment for this

wemeetagain (commented):
structure looks good :+1: mainly left comments about.. adding more comments :sweat_smile:

src/index.ts
tuyennhv: yeah that comment was for "exclude" strategy, will fix it

tuyennhv: the comment does not go with the code, will fix it 👍

src/index.ts
tuyennhv: > newMeshPeers could also be a Set

👍 

> Maybe try making candidateMeshPeers a Set

for each heartbeat we want to shuffle the peers (the random context in `getRandomGossipPeers`) so we need an ordered collection. It sounds like the `LinkedList` implementation in lodestar would be suitable here as the `splice()` method is cheap and no need to reindex

tuyennhv: as commented above, will use LinkedList here as we need an ordered collections

src/utils/set.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Exclude up to `ineed` items from a set if item meets condition `cond`
+ */
+export function excludeItemsFromSet<T>(
```
wemeetagain, line 4: can you call these `removeItemsFromSet` and `removeFirstNItemsFromSet`?

src/utils/set.ts
```diff
@@ -0,0 +1,30 @@
...
+  superSet: Set<T>,
+  ineed: number,
+  cond: (peer: T) => boolean = () => true
+): Set<T> {
+  let count = 0
```
wemeetagain, line 9: I think you can just use `subset.size`

wemeetagain (approved):
:tada:

PR #270, opened 2022-06-08
feat: add msgIdToStrFn option

### LINKED ISSUE #269: Improve messageIdToString
**Motivation**
- right now `messageIdToString` uses string concatenation which is not memory efficient and it may trigger `gc` more frequently

Screen Shot 2022-06-08 at 11 37 56

- This was taken from a running node where it shows string concatenation
Screen Shot 2022-06-08 at 11 34 18

see the same issue here https://github.com/ChainSafe/lodestar/issues/3446

**Description**

- we should be able to switch to `Bufffer.toString('base64')` instead
- write unit test and perf test to make sure
- should partly improve the heartbeat performance #256

tuyennhv:
**Motivation**
- Lodestar would like to use `Buffer.from` to convert Uint8Array message id to string as it's more efficient as found in https://github.com/ChainSafe/lodestar/issues/3446

**Description**
- Provide `msgIdToStrFn` option so that application can provide its own implementation
- Closes #269

### DIFF
--- a/src/message-cache.ts
+++ b/src/message-cache.ts
@@ -1,6 +1,6 @@
 import type { RPC } from './message/rpc.js'
-import type { MsgIdStr, PeerIdStr, TopicStr } from './types.js'
-import { messageIdFromString, messageIdToString } from './utils/index.js'
+import type { MsgIdStr, PeerIdStr, TopicStr, MsgIdToStrFn } from './types.js'
+import { messageIdFromString } from './utils/messageIdToString.js'
 
 export interface CacheEntry {
   msgId: Uint8Array
@@ -26,6 +26,8 @@ interface MessageCacheEntry {
 export class MessageCache {
   msgs = new Map<MsgIdStr, MessageCacheEntry>()
 
+  msgIdToStrFn: MsgIdToStrFn
+
   history: CacheEntry[][] = []
 
   /**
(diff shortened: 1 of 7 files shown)


mpetrunic:
isn't this feature?

wemeetagain (changes requested):
Structure lgtm

src/message-cache.ts
```diff
@@ -26,20 +26,24 @@ interface MessageCacheEntry {
...
    */
   constructor(
     /**
-     * he number of indices in the cache history used for gossiping. That means that a message
+     * The number of indices in the cache history used for gossiping. That means that a message
      * won't get gossiped anymore when shift got called `gossip` many times after inserting the
      * message in the cache.
      */
     private readonly gossip: number,
+    msgIdToStrFn: MsgIdToStrFn,
```
dapplion, line 43: Why put this option in the middle? Feels better as last parameter

dapplion:
Is it better to inject Buffer.from at Lodestar level or we can make assumptions here? I would imagine that anyone using gossipsub at NodeJS would benefit from such a change. @mpetrunic @wemeetagain is it a problem to multiplex NodeJS / browser code at the library level?

mpetrunic:
> Is it better to inject Buffer.from at Lodestar level or we can make assumptions here? I would imagine that anyone using gossipsub at NodeJS would benefit from such a change. @mpetrunic @wemeetagain is it a problem to multiplex NodeJS / browser code at the library level?

Generally, we can, but afaik whatever you do, if somebody uses webpack 5 will have to add buffer polyfill (or at least disable it manually). I don't imagine other users need that optimization, so it might be better to just pass the function on the lodestar level.

dapplion (approved):

wemeetagain (approved):

Issue #271, opened 2022-06-09
Improve mcache


tuyennhv:
- [ ] At each heartbeat, to get gossip message right now if there are n topics we have to loop through message ids n time, should do it in 1 loop
- [ ] Also we should only gossip validated messages (see https://github.com/libp2p/rust-libp2p/blob/0d3787ed04de8997d6b4d8c2522e412b6ec0f455/protocols/gossipsub/src/mcache.rs#L167)
- [x] ~~As we don't need random access, consider using LinkedList if it's faster to loop.~~ Update: LinkedList is not faster if only push then loop
- [ ] Consider caching the message id string and use it later when shift - check the cost of it, see below

<img width="925" alt="Screen Shot 2022-06-09 at 17 06 39" src="https://user-images.githubusercontent.com/10568965/172822433-44b23947-a444-487c-90ca-eb443ad1fd74.png">

- [x] ~~Add metrics to it~~ Update: we take `mcache.size()` per scraping metric already

this could be an improvement to #256

PR #274, opened 2022-06-10
feat: remove messageIdFromString

tuyennhv:
**Motivation**
- `messageIdFromString()` is not cheap, it takes 1.64 % of cpu time as in one of lodestar beacon node
Screen Shot 2022-06-10 at 15 33 31

- we always have Uint8Array message id in all of our flows, we convert to string and we have to convert it back which is a redundancy
- `mcache`: we receive string msgId in `put()`, convert it to Uint8Array to cache and convert it to string again in `shift()` which is another redundancy

**Description**
- Introduce MessageId `msgId: Uint8Array; msgIdStr: MsgIdStr` with the convention
  - msgId: Uint8Array
  - msgIdStr: string representation of msgId
  - messageId: new MessageId type containing both msgId and msgIdStr
- Cache MessageId in mcache
- remove `messageIdFromString()` completely
- Part of #271

**Memory cost**
- An mcache of 5000 items takes 1MB more, 20000 items take 4MB more
- As monitored in a lodestar node subscribing to all subnets, mcache size is only 400 so it's cheap enough

### DIFF
--- a/test/message-cache.spec.ts
+++ b/test/message-cache.spec.ts
@@ -5,6 +5,14 @@ import { MessageCache } from '../src/message-cache.js'
 import * as utils from '@libp2p/pubsub/utils'
 import { getMsgId } from './utils/index.js'
 import type { RPC } from '../src/message/rpc.js'
+import { MessageId } from '../src/types.js'
+
+const toMessageId = (msgId: Uint8Array): MessageId => {
+  return {
+    msgId,
+    msgIdStr: messageIdToString(msgId)
+  }
+}
 
 describe('Testing Message Cache Operations', () => {
   const messageCache = new MessageCache(3, 5, messageIdToString)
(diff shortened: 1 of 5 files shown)


test/message-cache.spec.ts
```diff
@@ -5,6 +5,14 @@ import { MessageCache } from '../src/message-cache.js'
 import * as utils from '@libp2p/pubsub/utils'
 import { getMsgId } from './utils/index.js'
 import type { RPC } from '../src/message/rpc.js'
+import { MessageId } from '../src/types.js'
+
+const toMessageId = (msgId: Uint8Array): MessageId => {
```
wemeetagain, line 10: this will need to be tweaked after #270

tuyennhv: no, we can use `messageIdToString()` function directly in test

wemeetagain (commented):
structure lgtm, just would like one comment to explain why `MessageId` is needed

dapplion (approved):

wemeetagain (approved):

PR #277, opened 2022-06-13
fix: only gossip validated messages

### LINKED ISSUE #271: Improve mcache

tuyennhv:
**Motivation**
- when getting gossip message ids, we should loop through mcache history 1 time
- we should only gossip validated messages

**Description**
- Loop through mcache history 1 time for all topics
- when `asyncValidation` option is undefined or false, messages are consider validated
- when publishing messages, they are considered validated

Closes #271

TODO: wait for https://github.com/ChainSafe/js-libp2p-gossipsub/pull/274

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -928,7 +928,8 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
         this.gossipTracer.deliverMessage(validationResult.messageId.msgIdStr)
 
         // Add the message to our memcache
-        this.mcache.put(validationResult.messageId, rpcMsg)
+        // if no validation is required, mark the message as validated
+        this.mcache.put(validationResult.messageId, rpcMsg, !this.opts.asyncValidation)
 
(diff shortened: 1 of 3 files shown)


wemeetagain (approved):
LGTM

PR #287, opened 2022-06-16
perf: improve metric performance

tuyennhv:
**Motivation**
- `gauge.inc()` takes 2.45% cpu time which is > 2 / 3 `runHeartbeat()` cpu time

**Description**
- `gauge.inc()` will hash the label objects and push to Map every time we call it
- so we we write some cached gauge metric class to only do that when scrape
  - gauge has no label: use `NoLabelCachedGauge`, this gives 7x improvement
  - gauge has 1 label: use `OneLabelCachedGauge`, this gives 1.2x improvement
  - gauge has 2 labels: use `TwoLabelCachedGauge`, this gives 2.5x improvement
- also remove unnecessary gauge metric

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -898,7 +898,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
    * May forward to all peers in the mesh.
    */
   private async handleReceivedMessage(from: PeerId, rpcMsg: RPC.Message): Promise<void> {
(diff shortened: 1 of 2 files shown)


dapplion:
> also remove unnecessary gauge metric

@tuyennhv Could you do this item in a separate PR?

About the optimization, is it best to do it here or on Lodestar?

tuyennhv:
closing as this PR was merged https://github.com/ChainSafe/js-libp2p-gossipsub/pull/433

PR #289, opened 2022-06-28
chore(deps)!: bump @libp2p deps

D4nte:
Bump enough @libp2p deps to fix master.


achingbrain:
When a stream is created by the muxer, it has no protocol, it's just a simple data channel. The stream is returned to the upgrader, which [runs the protocol negotiation algorithm](https://github.com/libp2p/js-libp2p/blob/master/src/upgrader.ts#L382) (currently [multistream-select](https://www.npmjs.com/package/@libp2p/multistream-select) but there will be others in future), after which the protocol will be set.  It's then handed back to the user.

This is all internal to libp2p so maybe it needs an interim type for the muxer to return which has no protocol, then the upgrader can convert it to the type with a protocol so the types will be nicer to use.

Effectively though, there will be a protocol set on the stream that is returned from `connection.newStream` since if protocol negotiation fails an error will be thrown.

> the protocol was moved from the Connection to the Stream

Not exactly - `Connection` instances don't have protocols since the connection has a muxer, which has lots of muxed streams, each of which have a protocol.

If you have a connection instance without a multiplexer, I guess you could could negotiate a protocol over it manually, but this is of little practical use because muxing is cheap and connections are expensive.

Anyway previously we had the negotiated protocol being returned alongside the stream which made it finicky to track, now it's a property of the stream itself.

achingbrain:
I think the current logic is fine, but if you're null-guarding on `.protocol` then adding a log line for the case where it's not set is probably good from a debugging point of view, even if it's unlikely to get hit in practice.

achingbrain (approved):

PR #293, opened 2022-06-29
feat: allow configuring stream limits

achingbrain:
The registrar defaults to only allowing 1x inbound and 1x outbound stream per protocol so allow configuring more than that.

I've plucked some defaults out of the air, these will need tuning to make sure they're not too restrictive or permissive.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -143,6 +143,9 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
   // Debug
   /** Prefix tag for debug logs */
   debugName?: string
+
+  maxInboundStreams?: number
+  maxOutboundStreams?: number
 }
 
 export interface GossipsubMessage {
@@ -318,6 +321,8 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
   readonly opts: Required<GossipOptions>
   private readonly metrics: Metrics | null
   private status: GossipStatus = { code: GossipStatusCode.stopped }
+  private maxInboundStreams: number
+  private maxOutboundStreams: number
 
   private heartbeatTimer: {
     _intervalId: ReturnType<typeof setInterval> | undefined
@@ -438,6 +443,9 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     this.score = new PeerScore(this.opts.scoreParams, this.metrics, {
       scoreCacheValidityMs: opts.heartbeatInterval
     })
+
+    this.maxInboundStreams = options.maxInboundStreams ?? 32
+    this.maxOutboundStreams = options.maxOutboundStreams ?? 128
   }
 
   getPeers(): PeerId[] {
@@ -483,7 +491,12 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     // Incoming streams
     // Called after a peer dials us
     await Promise.all(
-      this.multicodecs.map((multicodec) => registrar.handle(multicodec, this.onIncomingStream.bind(this)))
+      this.multicodecs.map((multicodec) =>
+        registrar.handle(multicodec, this.onIncomingStream.bind(this), {
+          maxInboundStreams: this.maxInboundStreams,
+          maxOutboundStreams: this.maxOutboundStreams
+        })
+      )
     )
 
     // # How does Gossipsub interact with libp2p? Rough guide from Mar 2022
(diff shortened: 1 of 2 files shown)


wemeetagain:
Isn't the max incoming/outgoing streams per-connection? So you'd only need 1 in both direction?

wemeetagain:
probably need to change the comments here: https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-registrar/src/index.ts#L15

I've been so confused by the errors I'm seeing running lodestar. Definitely related to this.

wemeetagain:
I think the incoming and outgoing limits should be the same no?

We're doing some funky thing where we open two streams (one inbound, one outbound) per peer, and we read on one and write to the other. I can't find anything in the spec that suggests that this is necessary though...
Since streams are duplexes, we should be able to handle both inbound and outbound traffic on one?

achingbrain:
> Since streams are duplexes, we should be able to handle both inbound and outbound traffic on one?

Yes, one stream should suffice.

> I think the incoming and outgoing limits should be the same no?

I'm not sure - my thinking was you wouldn't want a remote peer to connect to you then open a ton of streams and potentially fill up the incoming message buffer for each one, but you might want to open a ton of streams to a remote peer which is why I set it to be asymmetric.

> Isn't the max incoming/outgoing streams per-connection? So you'd only need 1 in both direction?

They are per-connection limits, yes but an application might have some weird requirement that needs multiple simultaneous streams?  I was definitely seeing IPFS test failures around IPNS over pubsub that were failing until I exposed the stream limits locally and upped the values.

> probably need to change the comments here

If they are confusing people then yes, most likely.

wemeetagain:
I think these defaults are too permissive. Gossipsub is a little different than many other protocols in that streams between peers are long-lived, and you only really need a single one. I'm not even sure what it means to have multiple gossipsub inbound or outbound streams for a single peer. (would published messages be pushed out to all streams? only the most recent?) 

Current behavior is we only currently 'track' two steams per peer (one inbound, one outbound).
If more inbound streams are created, they will overwrite the prior inbound stream:
https://github.com/ChainSafe/js-libp2p-gossipsub/blob/master/src/index.ts#L609 and
https://github.com/libp2p/js-libp2p-pubsub/blob/master/src/peer-streams.ts#L87

src/index.ts
```diff
@@ -143,6 +143,9 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
   // Debug
   /** Prefix tag for debug logs */
   debugName?: string
+
+  maxInboundStreams?: number
+  maxOutboundStreams?: number
```
wemeetagain, line 148: need jsdoc comments on these

src/index.ts
```diff
@@ -438,6 +443,9 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     this.score = new PeerScore(this.opts.scoreParams, this.metrics, {
       scoreCacheValidityMs: opts.heartbeatInterval
     })
+
+    this.maxInboundStreams = options.maxInboundStreams ?? 32
+    this.maxOutboundStreams = options.maxOutboundStreams ?? 128
```
wemeetagain, line 448:
```suggestion
    this.maxInboundStreams = options.maxInboundStreams
    this.maxOutboundStreams = options.maxOutboundStreams
```
maybe we just have no default (and just fall back to the upstream default)

src/index.ts
achingbrain: Done!

achingbrain:
https://github.com/libp2p/js-libp2p/pull/1301 increases the default stream limits so we can go with these for now, we can always tune further at a later date.

Talking to some of the libp2p folk, single, full-duplex or long-lived half-duplex gossipsub streams remove a bit of flexibility around things like low-power devices which may not be able to hold a connection open forever and may just want to periodically connect to send a message so I guess we should handle multiple incoming streams, also ones that only exist for the duration of one message, etc.

wemeetagain (dismissed):

src/index.ts
```diff
@@ -318,6 +330,8 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
   readonly opts: Required<GossipOptions>
   private readonly metrics: Metrics | null
   private status: GossipStatus = { code: GossipStatusCode.stopped }
+  private maxInboundStreams: number
+  private maxOutboundStreams: number
```
wemeetagain, line 334:
```suggestion
  private maxInboundStreams?: number
  private maxOutboundStreams?: number
```

wemeetagain (approved):

PR #298, opened 2022-07-08
feat: tweak stream handling

wemeetagain:
The ovararching problem with our previous approach is that we're too carefree about managing streams for this protocol.
Specific problems include:
- creating multiple duplicate outbound streams.
- relying solely on a topology to trigger an outbound stream.

Creation and tracking of new streams is tweaked:
- Add an outbound stream queue
This used to ensure that outbound streams are created sequentially.
This stops multiple duplicate streams from being created for a single peer.
- Attempt to create an outbound stream when an inbound stream is created.
This allows us to connect to peers that haven't yet been registered by the topology but want to connect to us.
- Remove `PeerStreams` usage in favor of managing inbound/outbound streams separately
`PeerStreams` works, but obscures some lingering issues. It's also quite inflexible and annoying to modify it since `PeerStreams` implements an interface in `@libp2p/interface-pubsub`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,10 +78,13 @@
     "@libp2p/peer-record": "^2.0.0",
     "@libp2p/pubsub": "^3.0.0",
     "@libp2p/topology": "^3.0.0",
+    "abortable-iterator": "^4.0.2",
     "denque": "^1.5.0",
     "err-code": "^3.0.1",
     "iso-random-stream": "^2.0.2",
+    "it-length-prefixed": "^7.0.1",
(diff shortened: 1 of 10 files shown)


mpetrunic (approved):
Have you already tried this in lodestar? Any improvements?

wemeetagain:
Yeah I've linked it against lodestar and it works well.
No more duplicate outbound streams.
No more connections with zero outbound streams.

Also seems like we're getting beacon blocks more on time? I'm seeing a smaller head drift.

PR #310, opened 2022-08-01
feat: switch back to protobufjs

tuyennhv:
**Motivation**
- protons is too slow to use, see https://github.com/ipfs/protons/issues/51

**Description**
- Switch back to protobufjs
- Copy generated files from v1.0.0
- Change `rpc.js` to `rpc.cjs`, new `rpc.js` to reexport `RPC` from `rpc.js`


mpetrunic:
@tuyennhv have you tried with new versions of protons?https://github.com/ipfs/protons/releases/tag/protons-v4.0.1
https://github.com/ipfs/protons/compare/protons-runtime-v2.0.1...protons-runtime-v2.0.2

should have performance optimizations

src/message/index.ts
```diff
@@ -1 +0,0 @@
-export * from './rpc.js'
```
wemeetagain, line 1: We still need this file for the subpath export (`import ... from '@chainsafe/libp2p-gossipsub/message'`)

tuyennhv: yeah this is exactly what I got when I build lodestar 👍

tuyennhv:
> @tuyennhv have you tried with new versions of protons?https://github.com/ipfs/protons/releases/tag/protons-v4.0.1
https://github.com/ipfs/protons/compare/protons-runtime-v2.0.1...protons-runtime-v2.0.2

@mpetrunic I've tried it, below is the result:
- current protons
```
rpc
    ✔ decode Attestation message                                           7089685 ops/s    141.0500 ns/op        -    2045205 runs   30.1 s
```
- new protons (v4.0) & protons-runtimes (v2.0)
```
rpc
    ✔ decode Attestation message                                           6807815 ops/s    146.8900 ns/op        -    1977187 runs   30.1 s
```
- protobufjs
```
rpc
    ✔ decode Attestation message using protobufjs                      1.315789e+8 ops/s    7.600000 ns/op        -   28030718 runs   30.0 s
```

=> there is no improvement made with new protons and it's still 20x slower than protobufjs

dapplion (approved):
LGTM!

wemeetagain (approved):

achingbrain:
> there is no improvement made with new protons and it's still 20x slower than protobufjs

Can you please share the benchmarks you're using to make these measurements?

tuyennhv:
> > there is no improvement made with new protons and it's still 20x slower than protobufjs
> 
> Can you please share the benchmarks you're using to make these measurements?

@achingbrain the benchmark is in https://github.com/ChainSafe/js-libp2p-gossipsub/issues/318

Issue #313, opened 2022-08-04
v03.5.0: Too many topic peers


tuyennhv:
**Description**
There are only 21 peers
<img width="842" alt="Screen Shot 2022-08-04 at 12 25 13" src="https://user-images.githubusercontent.com/10568965/182769351-e56498b9-540f-4d56-acad-0b84469bae68.png">

But it showed a lot of topic peers

<img width="1591" alt="Screen Shot 2022-08-04 at 12 26 08" src="https://user-images.githubusercontent.com/10568965/182769429-0f91928d-eb78-4b21-8b71-e47351ad7233.png">

Quick debug
<img width="585" alt="Screen Shot 2022-08-04 at 12 28 28" src="https://user-images.githubusercontent.com/10568965/182769691-555f193c-7b9a-4e43-b1f1-93ce73c3a447.png">

tuyennhv:
there are too many connected peers too

<img width="1242" alt="Screen Shot 2022-08-04 at 16 58 17" src="https://user-images.githubusercontent.com/10568965/182819908-a2b14379-f907-4e54-a726-72d893f5f482.png">

looks like there are some peers disconnected without calling `GossipSub.removePeer()` properly

Expected:

<img width="1238" alt="Screen Shot 2022-08-04 at 17 00 32" src="https://user-images.githubusercontent.com/10568965/182820323-77051de0-ec68-4c99-9899-ec3f9f64d7d9.png">

wemeetagain:
It seems that this library relies on the registrar properly registering a peer to this protocol in order for both the onPeerDisconnected and onPeerConnected routines to be triggered.
So if we get an inbound-negotiated stream with a peer and libp2p doesn't successfully get a response from the identify protocol, then the registrar won't call onPeerDisconnected for that peer. 
Likewise, if we connect to a peer, but don't get a response from the identify protocol, then we won't initiate a gossipsub stream with that peer (unless that peer opens a gossipsub stream).

I think for the purposes of networks that do not necessarily support the identify protocol (eg: ethereum network) we need the following:
- A more robust way to ensure that all closed connections with peers result in proper cleanup in the gossipsub router (rather than assume the registrar will trigger cleanup on its own)
eg:
```ts
// in gossipsub
connectionManager.addEventListener('peer:disconnect', this.onPeerDisconnected.bind(this))
```
- A way to pass a negotiated gossip stream to the router (rather than assume the registrar will open one on its own)
eg:
```ts
[...]

PR #319, opened 2022-08-09
fix: handle peer:disconnect event from connection manager

tuyennhv:
**Motivation**
Sometimes `onPeerDisconnected()` is not called by the registrar and it leaves a lot of open streams even through the peer is disconnected

**Description**
Handle `peer:disconnect` from connection manager

part of #313

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -475,6 +475,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
   async init(components: Components): Promise<void> {
     this.components = components
     this.score.init(components)
+    this.components.getConnectionManager().addEventListener('peer:disconnect', this.onConnectionClosed.bind(this))
   }
 
   /**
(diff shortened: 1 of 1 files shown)


tuyennhv:
this should fix the issue that we keep pushing data to a closed raw stream
Screen Shot 2022-08-09 at 12 45 00

and it keeps accumulating inside the internal buffers of `pushable`

Screen Shot 2022-08-09 at 12 47 16

wemeetagain:
LGTM @achingbrain do you have any comments on this before we merge?

achingbrain:
Do you know under what circumstances onPeerDisconnected is not being called?

wemeetagain:
Take a look at this issue https://github.com/ChainSafe/js-libp2p-gossipsub/issues/313

Not entirely sure what is triggering it in practice, but just looking at the code, it can happen when an inbound gossip stream is established with a peer that doesn't support (or doesn't successfully complete) the identify protocol. Then that peer disconnects.

wemeetagain:
Discussed with @achingbrain , he will add more functionality upstream in js-libp2p to support this rather than make a change here.
When inbound or outbound protocol streams are negotiated, libp2p will make sure the protocol is registered to that peer in the protobook.

wemeetagain:
closing in favor of https://github.com/libp2p/js-libp2p/pull/1341

PR #334, opened 2022-08-24
fix: catch errored push of rpc

wemeetagain:
In #325 , pushing an rpc thru the outbound stream can now throw.

This error should be caught to avoid interrupting publishing and forwarding messages.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2111,7 +2111,21 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     }
 
     const rpcBytes = RPC.encode(rpc).finish()
-    outboundStream.push(rpcBytes)
+    try {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -2111,7 +2111,21 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
...
+
+      // if the peer had control messages or gossip, re-attach
+      if (ctrl) {
+        this.control.set(id, ctrl)
+      }
+      if (ihave) {
+        this.gossip.set(id, ihave)
+      }
+
+      return false
```
dapplion, line 2127: @wemeetagain instead of having to throw an error, why don't we do the "stream is full" check here? try / catch needlessly is expensive

PR #336, opened 2022-09-20
fix: update event type to extend `{ [s: string]: any }`

achingbrain:
The `Events` type has to extend `{ [s: string]: any }`. Hopefully one day TS will ship a typed `EventEmitter` and this won't be necessary any more.

Fixes a type error:

```
[10:45:49] tsc [started]
test/utils/events.ts:64:25 - error TS2344: Type 'Events' does not satisfy the constraint '{ [s: string]: any; }'.

64   emitter: EventEmitter<Events>,
                           ~~~~~~

  test/utils/events.ts:63:35
    63 export const awaitEvents = async <Events = GossipsubEvents>(
                                         ~~~~~~~~~~~~~~~~~~~~~~~~
    This type parameter might need an `extends { [s: string]: any; }` constraint.


Found 1 error in test/utils/events.ts:64
```

PR #337, opened 2022-09-21
feat: add `src` folder to package to enable access to source map

fryorcraken:
Useful when the project use tooling to access the source map.


fryorcraken:
CI failure as discussed in #336

wemeetagain:
can you rebase off master, build should be patched with #340

Issue #338, opened 2022-09-22
CI is broken


tuyennhv:
**Description**
Got this error in https://github.com/ChainSafe/js-libp2p-gossipsub/actions/runs/3095619454/jobs/5024578133

```
[02:01:55] tsc [started]
Error: src/index.ts(511,68): error TS2345: Argument of type 'import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@multiformats/multiaddr/dist/src/index").Multiaddr[]' is not assignable to parameter of type 'import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@libp2p/interface-connection/node_modules/@multiformats/multiaddr/dist/src/index").Multiaddr[]'.
  Type 'import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@multiformats/multiaddr/dist/src/index").Multiaddr' is not assignable to type 'import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@libp2p/interface-connection/node_modules/@multiformats/multiaddr/dist/src/index").Multiaddr'.
    Types of property 'encapsulate' are incompatible.
      Type '(addr: import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@multiformats/multiaddr/dist/src/index").MultiaddrInput) => import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@multiformats/multiaddr/dist/src/index").Multiaddr' is not assignable to type '(addr: import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@libp2p/interface-connection/node_modules/@multiformats/multiaddr/dist/src/index").MultiaddrInput) => import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@libp2p/interface-connection/node_modules/@multifo...'.
        Types of parameters 'addr' and 'addr' are incompatible.
[...]

PR #339, opened 2022-09-22
fix: update @multiformats/multiaddr to 11.0.0

achingbrain:
As far as I can see this is only used in an interface so the only change needed is to the `package.json` file.


tuyennhv (approved):
confirm this fixes issue #338

PR #340, opened 2022-09-23
fix: copy js files to dist manually

tuyennhv:
**Description**
- There is build issue
```
Error: R] No matching export in "dist/src/message/rpc.cjs" for import "default"

    dist/src/message/rpc.js:1:7:
      1 │ import cjs from "./rpc.cjs";
```
- root cause is
```
This file is considered to be an ECMAScript module because of the "export" keyword here:

    dist/src/message/rpc.cjs:1747:0:
      1747 │ export {};
```

**Description**
During build process, js files are automatically copied to `dist` folder and some how it added 1 extra line at the end
```
export {};
```
and it made `rpc.cjs` file an esm module

- The fix is to copy js files automatically

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "scripts": {
     "lint": "eslint --ext .ts src test",
     "release": "aegir release --no-types",
-    "build": "aegir build && cp src/message/rpc.d.ts dist/src/message",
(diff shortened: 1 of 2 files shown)


wemeetagain:
Hopefully we can get rid of this once we migrate to the new protons :see_no_evil:

wemeetagain (approved):

PR #345, opened 2022-09-28
feat: add batch mode to publish api

### LINKED ISSUE #344: Publish in batch
**Motivation**
There is a missed attestation issue described in https://github.com/ChainSafe/lodestar/issues/4600. When lodestar wants to publish a gossip message, it wants to spread to the network asap

**Description**
Right now the steps to publish a message are as below:
- Select peers to publish, this is usually about 20 peers (as in lodestar)
- For each peer:
  - Form a new rpc for that peer with cached control messages of that peer
  - Encode message to bytes using protobuf
  - Encode to length prefix
  - Send to raw stream

The proposal is to send in batch without checking cached control messages
- Select peers to publish
- Encode messages to bytes using protobuf, then encode length prefix
- Send that bytes data to all raw stream

**Something to consider**
- Is it against the spec @wemeetagain 
- Should we add a new api or add a new flag to `publish()` method? Main thing is not to introduce a breaking change
- Does it have any side effect to control messages? Since we have to forward messages very frequently, and we also have  heartbeat, I suppose "no" optimistically

tuyennhv:
**Description**
- Add batch mode to publish() api
- No need to consider the cached control messages
- Encode length prefixed data and same this same data to all outbound raw stream

Closes #344

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,7 @@ import { Logger, logger } from '@libp2p/logger'
 import { createTopology } from '@libp2p/topology'
 import type { PeerId } from '@libp2p/interface-peer-id'
 import { CustomEvent, EventEmitter } from '@libp2p/interfaces/events'
+import { encode } from 'it-length-prefixed'
 
 import { MessageCache } from './message-cache.js'
 import { RPC, IRPC } from './message/rpc.js'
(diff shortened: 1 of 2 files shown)


tuyennhv:
replaced by #480

PR #347, opened 2022-09-28
fix: sendRpc in a for loop

tuyennhv:
**Description**
- The `sendRpc()` function expect new rpc param every time we call it because it modifies that param. If we call it in a for loop, the control part could be accumulated from prior loops.
- The issue may happen in `forward` and `publish` scenario


mpetrunic:
"feat: fix"? 😂

dapplion (approved):
Removed `createGossipRpc` function completely

dapplion (approved):

PR #348, opened 2022-09-29
feat: allow only defined list of topics

dapplion:
Allow to provide a whitelist of known topics by the library consumer.

If `allowedTopics` is provided, gossipsub will drop all subscriptions and messages with topics not in the whitelist.


mpetrunic (commented):
Maybe a test should be added to cover this new functionality. Otherwise LGTM

tuyennhv (approved):

PR #349, opened 2022-09-29
feat: Add stream option to limit inbound message size

g11tech:
If the size is not provided `decode` from `it-length-prefixed` lib applies the ~4mb limit by default:
From npm docs: 
```
maxDataLength: If provided, will not decode messages whose data section exceeds the size specified, if omitted will use the default of 4MB.
```

bellatrix message sizes could be 10MB so this limit needs to be exposed via options so they can be set by lodestar

### DIFF
--- a/src/stream.ts
+++ b/src/stream.ts
@@ -10,6 +10,11 @@ type OutboundStreamOpts = {
   maxBufferSize?: number
 }
 
+type InboundStreamOpts = {
+  /** Max size in bytes for reading messages from the stream */
+  maxDataLength?: number
+}
+
 export class OutboundStream {
   private readonly pushable: Pushable<Uint8Array>
   private readonly closeController: AbortController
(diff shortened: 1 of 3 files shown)


src/stream.ts
```diff
@@ -54,11 +59,13 @@ export class InboundStream {
   private readonly rawStream: Stream
   private readonly closeController: AbortController
 
-  constructor(rawStream: Stream) {
+  constructor(rawStream: Stream, opts: InboundStreamOpts = {}) {
     this.rawStream = rawStream
     this.closeController = new AbortController()
 
-    this.source = abortableSource(pipe(this.rawStream, decode()), this.closeController.signal, { returnOnAbort: true })
+    this.source = abortableSource(pipe(this.rawStream, decode(opts)), this.closeController.signal, {
```
mpetrunic, line 66: Should we add some default maxDataLength value?

g11tech: we can default to bellatix 10 mb allthough decode itself defaults to 4mb, what do you say?

mpetrunic: if it has default limit, than it's fine

dapplion (dismissed):

g11tech:
I think i will need to fix the tests since the pubsub reference has changed, let me update and push

g11tech:
@dapplion ready for merge!

dapplion (approved):

wemeetagain (approved):

PR #350, opened 2022-10-03
feat: allow to customize decode RPC function

dapplion:
Lodestar has some specific security needs and would require a hardened version of RPC.decode. Customizing this function allows Lodestar to experiment without requiring constant upstream changes


tuyennhv (approved):

dapplion:
Not necessary with https://github.com/ChainSafe/js-libp2p-gossipsub/pull/352

PR #353, opened 2022-10-03
fix: flip conditional in piggyback control - typo from #347

dapplion:
- Fixes typo from https://github.com/ChainSafe/js-libp2p-gossipsub/pull/347 with did not include a negation in a conditional check

@wemeetagain and I believe this logic appears strange and may be flipped. Asked in relevant channels https://discord.com/channels/806902334369824788/942671304530747402/1026542001082077184


dapplion:
Merging to get master green again

Issue #356, opened 2022-10-05
Memory leak due to hash collision


tuyennhv:
**Description**
With a faster version of `fastMsgIdFn` (see https://github.com/ChainSafe/lodestar/pull/4639 and https://github.com/ChainSafe/lodestar/issues/4603), the cache size somehow keeps increasing

<img width="788" alt="Screen Shot 2022-10-05 at 20 36 23" src="https://user-images.githubusercontent.com/10568965/194073881-1ebd6cd8-444a-4012-8f57-a1f99e0c8e70.png">

<img width="819" alt="Screen Shot 2022-10-05 at 20 40 36" src="https://user-images.githubusercontent.com/10568965/194074749-2f51a470-2be1-4dcd-aa9c-c969ae9278f0.png">

heartbeat runs regularly which should prune the SimpleTimeCache regularly too

<img width="800" alt="Screen Shot 2022-10-05 at 20 37 15" src="https://user-images.githubusercontent.com/10568965/194074052-2ff8111c-10bf-464c-9be3-e9d8640244a4.png">

on a good node with the slower/current `fastMsgIdFn` using sha256, it shows less than 1/20 cache size and it does not grow forever

<img width="796" alt="Screen Shot 2022-10-05 at 20 38 38" src="https://user-images.githubusercontent.com/10568965/194074336-23da7484-7094-411c-baf3-e37da534b9ef.png">

tuyennhv:
<img width="584" alt="Screen Shot 2022-10-06 at 19 15 49" src="https://user-images.githubusercontent.com/10568965/194318829-ec0304c4-c978-415c-85ce-eff282080091.png">

we can see that there are a lot of entries have smaller `validUntilMs` than the 1st entry

this is because of how frequent the data collision we have with xx-hash 32, a latest entry was hashed to the same value of an existing entry and overwrite the `validUntilMs` there, then it's not possible for `prune` to does its job, reproducable with this unit test

```typescript
const testCases = [1e4, 1e5];
  for (const size of testCases) {
    it("should return consistent value " + size, async () => {
      const entries = new Map<number, number>();

      for (let i = 0; i < size; i++) {
        const data = crypto.randomBytes(32) as Uint8Array;
        const message = {data} as RPC.IMessage;
        const h = fastMsgIdFn(message);
        entries.set(h, i);
      }

      let i = 0;
      for (const [_, v] of entries.entries()) {
        expect(v).to.be.equal(i);
        i++;
[...]

PR #357, opened 2022-10-07
feat: implement putUnsafe() for time cache

tuyennhv:
**Motivation**
- There is memory leak issue in the case of `fastMsgIdFn` hash function data collision, which causes `validUntilMs` to be not in ascending order and `prune` function does not work well
- It's not only happen in #356, I think the current version of lodestar (v1.1.0) also has this issue that's why we have this pattern

Screen Shot 2022-10-07 at 15 38 12

In very bad case `fastMsgIdCache` could grow a lot.

**Description**
- Implement `putUnsafe()` function
- It's noticed that the `putUnsafe()` is 2x slower than `put()`

**TODO**
- [ ] Test in a node for some time


tuyennhv:
implemented in #358

PR #358, opened 2022-10-07
fix: TimeCache handle key collision to prevent leak

dapplion:
- @tuyennhv detected a potential leak due to bad prune logic in TimeCache https://github.com/ChainSafe/js-libp2p-gossipsub/issues/356

This doesn't show up often since we currently use in Lodestar sha256 function with 8 bytes so the chance of collision is exceedingly rare.

Closes https://github.com/ChainSafe/js-libp2p-gossipsub/issues/356


test/timeCache.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { expect } from 'aegir/chai'
+
+describe.only('utils / TimeCache', function () {
```
tuyennhv, line 3: remove `only`

tuyennhv: there is also an existing `time-cache.test.ts`

dapplion:
CI failed with 

```
Error: test/compliance.spec.ts(19,25): error TS2345: Argument of type 'import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@libp2p/interface-pubsub-compliance-tests/node_modules/@libp2p/components/dist/src/index").Components' is not assignable to parameter of type 'import("/home/runner/work/js-libp2p-gossipsub/js-libp2p-gossipsub/node_modules/@libp2p/components/dist/src/index").Components'.
  Types have separate declarations of a private property 'peerId'.
Error: test/utils/create-pubsub.ts(23,5): error TS2741: Property 'acceptIncomingConnection' is missing in type 'MockConnectionManager' but required in type 'ConnectionManager'.
```

wemeetagain (approved):

PR #360, opened 2022-10-10
fix!: remove @libp2p/components

achingbrain:
`@libp2p/components` is a choke-point for our dependency graph as it depends on every interface, meaning when one interface revs a major
`@libp2p/components` major has to change too which means every module depending on it also needs a major.

Switch instead to constructor injection of simple objects that let modules declare their dependencies on interfaces directly instead of
indirectly via `@libp2p/components`

Refs libp2p/js-libp2p-components#6

BREAKING CHANGE: modules no longer implement `Initializable` instead switching to constructor injection


achingbrain:
The latest `@libp2p/components` is breaking because the transport and upgrader interfaces have changed.  This obviously has nothing to do with pubsub so is not a great state of affairs - there's [some discussion here](https://github.com/libp2p/js-libp2p-components/issues/6) about how we can improve this and allow ourselves to make breaking changes to interfaces that then mean we don't need to upgrade the entire ecosystem.

achingbrain:
@wemeetagain looks like the e2e test timed out in CI?

src/index.ts
```diff
@@ -361,7 +375,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements Initiali
     cancel: () => void
   } | null = null
 
-  constructor(options: Partial<GossipsubOpts> = {}) {
+  constructor(components: GossipSubComponents, options: Partial<GossipsubOpts> = {}) {
```
dapplion, line 378: This is a breaking change?

dapplion (approved):

src/index.ts
achingbrain: Yeah, as noted in the PR title & description

wemeetagain (approved):

PR #369, opened 2022-11-03
feat: remove unnecessary conversion from Multiaddr to IP

dapplion:
Gossipsub currently polls connection manager frequently to figure what IPs each peer connection has. This is very inneficient and causes performance issues in Lodestar

- Similar to https://github.com/ChainSafe/discv5/pull/214

Instead js-libp2p should follow a similar pattern to rust-libp2p, which injects an `AddressChange` event into protocols. See https://github.com/sigp/rust-libp2p/blob/a3dec471c046127e8e0d88bc89a881affb56a76c/protocols/gossipsub/src/behaviour.rs#L3282

@achingbrain is js-libp2p emitting a similar event already? Where could this be added if not?


wemeetagain:
Good change, here's the event you're looking for: https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-peer-store/src/index.ts#L209

dapplion:
:heavy_check_mark: looks good! Can't approve since it's my PR

wemeetagain (approved):

PR #370, opened 2022-11-03
fix: add .js extension to imported file

AlexMesser:
Added `.js` extension to imported file since compiler gives an error:

 `Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './rpc.js'?` 

with `"moduleResolution"` option set to `"NodeNext"`


wemeetagain (approved):
thanks

Issue #380, opened 2022-12-06
Tag mesh peers on graft, remove tags on prune


achingbrain:
To prevent the connection manager pruning connections to high-value peers, modules that need connections to certain peers should tag them to ensure they don't get culled by the connection manager when connection limits have been specified and reached.

When this module grafts mesh peers for a given topic, it should use the [peerStore.tagPeer](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-peer-store/src/index.ts#L255) method to tag them with a value:

```js
libp2p.peerStore.tagPeer(peerId, 'gossipsub-mesh-peer', {
  value: 100 // value should be 0-100
})
```

When they are pruned from a mesh [peerStore.unTagPeer](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-peer-store/src/index.ts#L256) should be used:

```js
libp2p.peerStore.unTagPeer(peerId, 'gossipsub-mesh-peer')
```

Refs: https://github.com/libp2p/js-libp2p/issues/369

dapplion:
Do other implementations have this feature? I don't recall seeing it. Speaking for Lodestar we have more complex logic to decide the value of the peer so this feature won't be useful at all.

achingbrain:
> Do other implementations have this feature? I don't recall seeing it.

See the referenced issue in the OP for context.

> Speaking for Lodestar we have more complex logic to decide the value of the peer 

If that's the case lodestar may wish to tag high value peers as well.

dapplion:
> If that's the case lodestar may wish to tag high value peers as well.

If I understand this feature is relevant because the connection manager may choose to disconnect peers at will right? Can consumers completely disable that so peer management is done 100% by the consumer?

achingbrain:
Yes, just leave `maxConnections` as the default (`Infinity`).

PR #383, opened 2022-12-14
feat: graft/prune events and mesh peer tagging

maschad:
- Add `'gossipsub:graft'` and `'gossipsub:prune'` events with the following detail:
```ts
export interface MeshPeer {
  peerId: string
  topic: string
  direction: Direction
}
```
- Add new configuration option `tagMeshPeers` (default to `true`) which will tag the mesh peers in the peer store (This may be used in turn by the connection manager to rank connections, eg for disconnection purposes)
- Closes https://github.com/ChainSafe/js-libp2p-gossipsub/issues/380


src/index.ts
```diff
@@ -1,31 +1,40 @@
...
-  createPeerScoreParams,
-  createPeerScoreThresholds,
-  PeerScoreStatsDump
-} from './score/index.js'
-import { IWantTracer } from './tracer.js'
-import { SimpleTimeCache } from './utils/time-cache.js'
+  Message,
+  PublishResult,
+  PubSub,
+  PubSubEvents,
```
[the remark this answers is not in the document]
maschad: Sorry about that, that's my vscode [Organize Imports](https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_68.md#group-aware-organize-imports) feature, I don't know if this is something worth considering in the future using some sort of [import order](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md)

wemeetagain:
comment from @dapplion
> Hey! I strongly oppose merging the peer tagging feature before having strong guarantees this won't have a performance penalty.

maschad:
> comment from @dapplion
> 
> > Hey! I strongly oppose merging the peer tagging feature before having strong guarantees this won't have a performance penalty.

Following a suggestion from @achingbrain we could do the tagging / untagging after the sending an rpc object to a peer (e.g. https://github.com/ChainSafe/js-libp2p-gossipsub/blob/master/src/index.ts#L1241) to unblock grafting / pruning operations, this should avoid the performance penalty. 

what do you think @dapplion ?

achingbrain:
> this should avoid the performance penalty

I think it's important to establish a baseline measurement to ensure we have a basis for comparison of different solutions before we make architectural decisions predicated on performance impact.

@dapplion do you have a specific metric in mind that changes here must not affect?  If you do, is there a benchmark that we can use as the baseline measurement?  If not, can one be created?

dapplion (changes requested):
As argued in https://github.com/libp2p/js-libp2p/issues/369#issuecomment-1386379491 I do not believe this feature is understood enough to be a net positive in js-libp2p. Collapsing a multidimensional score of "peer usefulness" into a single number can lead to unintended consequences. Note that Go and Rust gossipsub implementations do not use tagging.

maschad:
> As argued in [libp2p/js-libp2p#369 (comment)](https://github.com/libp2p/js-libp2p/issues/369#issuecomment-1386379491) I do not believe this feature is understood enough to be a net positive in js-libp2p. Collapsing a multidimensional score of "peer usefulness" into a single number can lead to unintended consequences. Note that Go and Rust gossipsub implementations do not use tagging.

The [go-libp2p-gossipsub implementation uses a tagTracer](https://github.com/libp2p/go-libp2p-pubsub/blob/3d2eab35722efeab8742ece94d2a2b2cc700f751/tag_tracer.go#L58) which applies tags to peer connections based on their behaviour, leveraging the [TagPeer methods from the connection manager implementation](https://github.com/libp2p/go-libp2p/blob/1ad0a50be62b1e00388b9d85fc997d66e8910c9d/core/connmgr/manager.go#L35-L38)

Peers are tagged based on the following criteria:
 - Directly connected peers are tagged with `GossipSubConnTagValueDirectPeer` (default 1000).
 -  Mesh peers are tagged with a value of `GossipSubConnTagValueMeshPeer` (default 20).
 If a peer is in multiple topic meshes, they'll be tagged for each.
- For each message that we receive, we bump a delivery tag for peer that delivered the message first (using the `GossipSubConnTagBumpMessageDelivery` variable)
- The delivery tags have a maximum value, GossipSubConnTagMessageDeliveryCap, and they decay at a rate of `GossipSubConnTagDecayAmount / GossipSubConnTagDecayInterval`.

So perhaps we need to expand this PR to accommodate those different types of tags and penalties when adjusting a [peer-score](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/master/src/score/peer-score.ts#L1) . What do you think @achingbrain @wemeetagain  ?

dapplion:
What happened to the diff? +14,000 -42,000

maschad:
> What happened to the diff? +14,000 -42,000

This is just for testing purposes which is why [the commit is labelled DNM](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/383/commits/57ab5173bb86b5ac029e0b6af78e2e7936a4ba7f)

maschad:
After deploying this on Feb 7th and observing on our larger cluster I have some observations. In summary

**Pros**
- Improved throughput
- More stable peer count
- better gossip scores

**Cons**
- Increased heap usage
- slightly increased CPU usage

for more details:


The screenshots are in order of **unstable** (`unstable-lg1k-hzax41`)  then  **feat2** (`feat2-lg1k-hzax41`) respectively.

## Gossip score (avg / min / max)

### Unstable
mpkatquRz7qf5JaOLWyCguEPtHmpn0f9mwn5wZIpZDufpd9SWeGoZYndMD4e_riQmX87hrDWiTR9WO5Mjb26_G9c4r_cKnGCdHI9eJk14qwDWiGmIyR8KsbsZ3Xp

### Feat2
pNyZLxTOYl6uvUCmxxu2Fb_c-JOI3_6rzW4Yia-a3oCbMQ9jmaaXj45SEr7A31V4J3brrYSgOE6FqAt0HmsdcwUh9MLuvdvVNpFtJcHh9KHarr4mOVGNJd59AcaG

We can observe that unstable on average had more low scores that deviated from the average than feat2 although the variance is not significant over the last 6 days.

## Peer count per gossip score threshold

### Unstable

CXkgPCLO-tjgjtwMvWacxCXNTQhHB8FypaXCqLkrkN--8zXlvXkIMq8ABlDE8hVPqhJfGLLgwG3JmkwFIkh-28ZaAFXWmzBUAS_nuN4G0WTsp4C5-4qnOd_o2jFz


### Feat2

7ZJfIZ5YUgIQIx2vd6fhavX6FVWBOqsM3ZniItNcz-Se-gU8MB3PAC46G54ekViEbuk1X0z_zNlDJBRdISiMhIPGlc1JvFIXmn4S2IxcJGLHjvcbI8stsQvZrSmf

We can observe that on feat2  there is substantially less variance over the last 6 days than unstable (around 9 at max) as opposed to spikes of 50 peers on unstable , the peer count remained relatively stable in feat2

## Process CPU user seconds

### Unstable

h-_dlymqO7Zt9gchY6uYYLvugATKt_xDFpuMPp2jCyWfHhMdIecHFWelgwVR2bPM-3EDylK0Ry7UrLQSGZCOMjwkhDPgHhE6bk9-j6ONDHh8QgdRKsmkUMqxqTEN


### Feat2

YNZ55ZY52xVhlMff5SVvRbkNcX3opHFg_am9sznj9av-RHmhmp4Jl9W92TQ49qXkH05wVa2OTeA2uXGsy-zh5wxhr4E19nWb8cQo0XGFRtsAk-jP2TpOgyzG4Nti

Slightly more on average CPU usage with feat2


## Process Heap Bytes

### Unstable

Vhd7UczCMxa8s9bTqF35Qt5DpaTBbNDPBMxpoBdztD60oslOIqSLoshEBAHyymNzcFV_oZklaXWA5RUJlZY1mU_osdplYCN6RMVZps_aCY5CkBvRX4XeZm6FYaJo

### Feat2

CZhDAWy-ZSry8XyzI89hYkUV-muWLlLCID52WjxcZPEu3wEYbcOYSF7emgFXTwqSw7MdA8IsAdvtn8dapR2Q1YxiDmnX6hC1pzVxzIn0jvozuv5eMuqNEnb4C-zn

There is consistently higher heap usage with feat2 with an average higher than unstable’s max.

## Received messages per second

### Unstable

Ss1gdZUQxxrK8K2a-ynX0LkXt9mdEFfNRiqzQznWVxUWiWxLGTrSLc0ylZc8et1mxT-jowyP2542Uz9QVi-WVrxG6etnR6XgZvVHySSbvR7E3XcpyfwR4P4P-ku2

### Feat2

ca1bdqBRKRYa5Z19uMabnBKThhHAWVt6iVd1IysJWmWPitPDVtGINJ8ZOe8A6nbe80Wn7CvecFBmd-t59UeSanhzctm3B5dLVFp1Mn_hexfdf1VbjcsTykRUBnnj

On average feat2 is receiving more messages per second over the last 6 days.

Would be interested to hear your thoughts @dapplion @wemeetagain @tuyennhv

maschad:
Just sending a reminder @dapplion @tuyennhv

dapplion:
> Just sending a reminder @dapplion @tuyennhv

What specific input do you need from us?

maschad:
> > Just sending a reminder @dapplion @tuyennhv
> 
> What specific input do you need from us?

I wanted to know if you thought these fluctuations were significant enough to be warranted as a performance penalty otherwise we could go ahead with this PR.

p-shahi:
Update: this is going to be redeployed by @wemeetagain and double checked.

p-shahi:
@wemeetagain :
Having some difficulties verifying this in prod. In production, several major versions behind (gossipsub) - can't easily deploy to a pre 0.45 libp2p released Lodestar
After verifying 0.45 works w.o regressions, this can be tested.

maschad:
@wemeetagain is going to share the metrics that illustrate below performant behaviour, but he would also like to have this functionality configurable in the interim given that connections are managed differently in Lodestar.

maschad:
Next steps on this, I will make this feature enabled where it will default to on, and Lodestar can have the option of disabling it.

dapplion:
diff is still huge

maschad:
I have reverted the `package-lock.json` which should reduce the diff

wemeetagain (changes requested):
This currently only adds / removes tags when we're receiving inbound GRAFT / PRUNE msgs.
Tagging / untagging also needs to be done when we're sending GRAFT / PRUNE msgs.

Example where this is a problem.
A peer grafts us, we tag it.
But they're a poorly performing peer. We prune them, but we don't untag them. So that peer never gets disconnected and replaced with a healthy peer.

maschad:
> Tagging / untagging also needs to be done when we're sending GRAFT / PRUNE msgs.

To my understanding is that's what is being done here no?

maschad:
I refactored slightly to place the tagging function in [`sendGraft`](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/383/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R2224-R2233) to make it more apparent.

wemeetagain:
Peers can be grafted on one topic and pruned on another topic.
This feature should take that into account.

Right now, if a peer is grafted onto topic A and B, but then pruned from topic A, their tag value goes away completely.

maschad:
Using the `topicID` as the tag key should solve this, pushed changes in 78a3669

wemeetagain:
Can you add a test that shows that the peer store is properly cleared of tags when gossipsub is shut down?

maschad:
Updated @wemeetagain mind you the value is set to `0` not that is removed as referenced in https://github.com/ChainSafe/js-libp2p-gossipsub/issues/380#issue-1479795857

wemeetagain:
> Updated @wemeetagain mind you the value is set to 0 not that is removed as referenced in https://github.com/ChainSafe/js-libp2p-gossipsub/issues/380#issue-1479795857

I don't see the reference to the tag being set to 0 in that comment

maschad:
My bad, I had misinterpreted the comment that says tag values should be between 0-100 to imply that for all cases, I realize in libp2p when untagging we set the metadata to `undefined` I have done so in c117cf9

maschad:
@wemeetagain [the test](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/383/files#diff-04d01547d20121630b2cd82cc5ea892c0366ff50769729e7cdcb699f0ae1c0cbR104) that actually checks for the tagging is flakey but I am not sure why, could it be that the peerStore is being cleared somewhere else?

src/index.ts
```diff
@@ -1471,7 +1471,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     const now = Date.now()
     let doPX = this.opts.doPX
 
-    graft.forEach(({ topicID }) => {
+    graft.forEach(async ({ topicID }) => {
```
dapplion, line 1474: Having to allocate a promise for every single graft packet for this feature is not ideal

maschad: Why is that?

dapplion: allocating promises has a non trivial cost in memory specially. We receive many grafts, so having to pay that cost for a feature that (at least) we don't use worsens performance for everyone. Ideally we want zero cost disabled features. Also returning promises to forEach is weird, as those are not handled.

Why don't you just do
```ts
for (const {topicID} of graft) {
```

maschad:
I've also updated `package-lock.json` as npm was reporting critical vulnerabilities.

src/index.ts
```diff
@@ -1856,15 +1861,27 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
...
+      this.sendGraft(id, topic)
 
       // rust-libp2p
       // - peer_score.graft()
       // - Self::control_pool_add()
       // - peer_added_to_mesh()
     })
+
+    if (this.opts?.taggingEnabled ?? false) {
+      Array.from(toAdd).map(async (id) => {
```
wemeetagain, line 1875: forEach or for-of

wemeetagain: probably for-of is better, then no need to create an array

maschad: then I would have to make this whole method `async` and the motivation was to move away from unnecessary promises. i.e. promises would only be created for consumers who enable tagging.

dapplion: What if `peerStore.merge` throws?

achingbrain:
libp2p triage update -

- Error handling to be improved
- Then review
- Then merge!

src/index.ts
wemeetagain: This should be refactored to a for-of

wemeetagain (dismissed):

src/index.ts
maschad: > then I would have to make this whole method `async` and the motivation was to move away from unnecessary promises. i.e. promises would only be created for consumers who enable tagging.

src/index.ts
```diff
@@ -1474,10 +1479,30 @@ export class GossipSub extends TypedEventEmitter<GossipsubEvents> implements Pub
     const now = Date.now()
     let doPX = this.opts.doPX
 
+    if (this.opts?.taggingEnabled ?? false) {
```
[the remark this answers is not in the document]
maschad: I think this can ge done in a follow up PR if necessary considering not only is this an optional feature but it one immediate use case is where it can be used prune a connection via the connection manager in where that connection should not be gossiped to, in the case where it is enabled.

wemeetagain:
@achingbrain can you review this? I'm happy with it.

achingbrain (approved):
LGTM, small nit with the hard-coded tag value but it's not essential.

PR #385, opened 2022-12-14
fix: tracer to track delivered message if duplicate

tuyennhv:
**Motivation**
- To investigate broken promises in https://github.com/ChainSafe/lodestar/issues/4818
- Consider this scenario
  - 2 messages have same fastMsgId but different msgIds (collision)
  - node sends IWANT control message and add a promise
  - message 1 comes and being tracked in fastMsgId
  - message 2 (as a response to IWANT) comes, since it has same fastMsgId it's a duplicate we didn't mask as "message delivered" in tracer, and it looks like the other node did not send us the requested messages

**Description**
- call `tracer. deliverMessage()` in the case of message duplicate

part of https://github.com/ChainSafe/lodestar/issues/4818

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1093,6 +1093,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
       case MessageStatus.duplicate:
         // Report the duplicate
         this.score.duplicateMessage(from.toString(), validationResult.msgIdStr, rpcMsg.topic)
+        this.gossipTracer.deliverMessage(validationResult.msgIdStr)
         this.mcache.observeDuplicate(validationResult.msgIdStr, from.toString())
         return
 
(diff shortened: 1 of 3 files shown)


dapplion (changes requested):
Can you add metrics to verify this hypothesis? 
- Add a second parameter to `deliverMessage(msg, isDuplicate: boolean)`
- In that function, if promise resolved and isDuplicate then `this.metrics.iwantPromiseResolvedFromDuplicate.inc(1)`

tuyennhv:
thanks @dapplion for the metric, it shows a lot of promises could be resolved by duplicate messages (this is from a mainnet node subscribing to all subnets)

Screen Shot 2022-12-15 at 10 06 02

it could reduce 80% 90% of broken promises (this was only tested 3h but we could trust as these 2 metrics match each other)

Screen Shot 2022-12-15 at 10 07 23

dapplion (approved):

Issue #367, opened 2022-10-28
Add per publish call allowPublishToZeroPeers option


dapplion:
For a specific publish() call allow to set allowPublishToZeroPeers to suppress such error.

PR #395, opened 2023-01-08
feat: added allowPublishToZeroPeers as optional param to publish function

maschad:
Closes https://github.com/ChainSafe/js-libp2p-gossipsub/issues/367

The default for allowPublishToZeroPeers will be false but a user can optionally call with allowPublishToZeroPeers set to true to suppress InsufficientPeers errors

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,7 +52,8 @@ import {
   DataTransform,
   rejectReasonFromAcceptance,
   MsgIdToStrFn,
-  MessageId
+  MessageId,
+  PublishOpts
 } from './types.js'
 import { buildRawMessage, validateToRawMessage } from './utils/buildRawMessage.js'
 import { msgIdFnStrictNoSign, msgIdFnStrictSign } from './utils/msgIdFn.js'
@@ -1995,7 +1996,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
    *
    * For messages not from us, this class uses `forwardMessage`.
    */
-  async publish(topic: TopicStr, data: Uint8Array): Promise<PublishResult> {
+  async publish(topic: TopicStr, data: Uint8Array, opts?: PublishOpts): Promise<PublishResult> {
     const transformedData = this.dataTransform ? this.dataTransform.outboundTransform(topic, data) : data
 
     if (this.publishConfig == null) {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1995,7 +1995,13 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
    *
    * For messages not from us, this class uses `forwardMessage`.
    */
-  async publish(topic: TopicStr, data: Uint8Array): Promise<PublishResult> {
+  async publish(
+    topic: TopicStr,
+    data: Uint8Array,
+    publishOptions: Partial<GossipsubOpts> = {
```
dapplion, line 2001: Using GossipsubOpts here is extremely broad, just allow `allowPublishToZeroPeers` and nothing else. You don't need to default to false, that what will happen anyway

maschad: Understood. I had made it more generic in case we may want to pass other params such as `emitSelf` in the future.

src/index.ts
```diff
@@ -1995,7 +1995,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
    *
    * For messages not from us, this class uses `forwardMessage`.
    */
-  async publish(topic: TopicStr, data: Uint8Array): Promise<PublishResult> {
+  async publish(topic: TopicStr, data: Uint8Array, allowPublishToZeroPeers = false): Promise<PublishResult> {
```
dapplion, line 1998:
```suggestion
  type PublishOpts {
    allowPublishToZeroPeers?: boolean
  }

  async publish(topic: TopicStr, data: Uint8Array, opts?: PublishOpts): Promise<PublishResult> {
    const allowPublishToZeroPeers = opts?.allowPublishToZeroPeers ?? this.opts.allowPublishToZeroPeers;
    
    ...
    
    if (
      tosend.size === 0 &&
      !allowPublishToZeroPeers
      !willSendToSelf
    ) {
      throw Error('PublishError.InsufficientPeers')
  }
```

maschad: I added the `PublishOpts` type to the [types](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/395/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR75) file.

dapplion (dismissed):

maschad:
Added a [test](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/395/files#diff-04d01547d20121630b2cd82cc5ea892c0366ff50769729e7cdcb699f0ae1c0cbR84)

wemeetagain (approved):
LGTM thanks @maschad

Issue #272, opened 2022-06-09
Allow to publish same message to multiple topics


dapplion:
Ethereum beacon chain requires to publish the same SyncCommitteeSignature to multiple topics on small networks.

Currently gossipsub prevents doing that by throwing `PublishError.Duplicate`

https://github.com/ChainSafe/js-libp2p-gossipsub/blob/17c25a1b4af1446615c42bdb2ed2f208b3f6a5ad/src/index.ts#L1858

Signature of `async publish(topic: TopicStr, data: Uint8Array)` could be changed to `topics: TopicStr[]` to publish to multiple topics. Or add an option to disable checking for duplicate messages

PR #404, opened 2023-02-14
feat: allow ignoring PublishError.Duplicate

### LINKED ISSUE #402: New option for publish api to ignore duplicate
**Description**
- Right now when lodestar publishes a block, builder may already published it and gossipsub throw `Duplicate` error in `publish` api
- I'd like to add `ignoreDuplicate` (or probably `allowDuplicate`?) option to publish api to return early and ignore the error

this is the root cause of https://github.com/ChainSafe/lodestar/issues/5140

maschad:
Closes https://github.com/ChainSafe/js-libp2p-gossipsub/issues/402

When `ignoreDuplicatePublishError` is set, instead of throwing with `PublishError.Duplicate`, simply return.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2022,7 +2022,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
       // This message has already been seen. We don't re-publish messages that have already
       // been published on the network.
       if (ignoreDuplicateMessages) {
-        this.metrics?.onIgnoreDuplicateMsg(topic)
+        this.metrics?.onIgnorePublishedDuplicateMsg(topic)
         return { recipients: [] }
       }
       throw Error('PublishError.Duplicate')
--- a/src/index.ts
+++ b/src/index.ts
@@ -110,7 +110,7 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
   /** Do not throw `InsufficientPeers` error if publishing to zero peers */
   allowPublishToZeroPeers: boolean
   /** Do not throw `PublishError.Duplicate` if publishing duplicate messages */
-  allowPublishDuplicates: boolean
+  ignoreDuplicateMessages: boolean
   /** For a single stream, await processing each RPC before processing the next */
   awaitRpcHandler: boolean
   /** For a single RPC, await processing each message before processing the next */
@@ -2016,11 +2016,14 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     const msgIdStr = this.msgIdToStrFn(msgId)
 
     // Current publish opt takes precedence global opts, while preserving false value
-    const allowPublishDuplicates = opts?.allowPublishDuplicates ?? this.opts.allowPublishDuplicates
+    const ignoreDuplicateMessages = opts?.ignoreDuplicateMessages ?? this.opts.ignoreDuplicateMessages
 
-    if (this.seenCache.has(msgIdStr) && !allowPublishDuplicates) {
+    if (this.seenCache.has(msgIdStr)) {
       // This message has already been seen. We don't re-publish messages that have already
       // been published on the network.
+      if (ignoreDuplicateMessages) {
+        return { recipients: [] }
+      }
       throw Error('PublishError.Duplicate')
     }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -109,6 +109,8 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
   asyncValidation: boolean
   /** Do not throw `InsufficientPeers` error if publishing to zero peers */
   allowPublishToZeroPeers: boolean
+  /** Do not throw `PublishError.Duplicate` if publishing duplicate messages */
+  allowPublishDuplicates: boolean
   /** For a single stream, await processing each RPC before processing the next */
   awaitRpcHandler: boolean
   /** For a single RPC, await processing each message before processing the next */
@@ -2013,7 +2015,10 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     const msgId = await this.msgIdFn(msg)
     const msgIdStr = this.msgIdToStrFn(msgId)
 
-    if (this.seenCache.has(msgIdStr)) {
+    // Current publish opt takes precedence global opts, while preserving false value
+    const allowPublishDuplicates = opts?.allowPublishDuplicates ?? this.opts.allowPublishDuplicates
+
+    if (this.seenCache.has(msgIdStr) && !allowPublishDuplicates) {
       // This message has already been seen. We don't re-publish messages that have already
       // been published on the network.
       throw Error('PublishError.Duplicate')
--- a/src/index.ts
+++ b/src/index.ts
@@ -110,7 +110,7 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
   /** Do not throw `InsufficientPeers` error if publishing to zero peers */
   allowPublishToZeroPeers: boolean
   /** Do not throw `PublishError.Duplicate` if publishing duplicate messages */
-  allowPublishDuplicateMessages: boolean
+  ignoreDuplicatePublishError: boolean
   /** For a single stream, await processing each RPC before processing the next */
   awaitRpcHandler: boolean
   /** For a single RPC, await processing each message before processing the next */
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -2013,7 +2015,10 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     const msgId = await this.msgIdFn(msg)
     const msgIdStr = this.msgIdToStrFn(msgId)
 
-    if (this.seenCache.has(msgIdStr)) {
+    // Current publish opt takes precedence global opts, while preserving false value
+    const allowPublishDuplicates = opts?.allowPublishDuplicates ?? this.opts.allowPublishDuplicates
+
+    if (this.seenCache.has(msgIdStr) && !allowPublishDuplicates) {
```
tuyennhv, line 2021: it's no use to publish again if there is a duplicate, just return here. In that case, I think we might want to go with a better name, `ignorePublishDuplicate` or any suggestions @wemeetagain ?

also let's add a gauge metric to track this with topic as a label

maschad: Thanks for clarifying @tuyennhv  I think I had misunderstood, I would say `ignoreDuplicateMessages` would be a more appropriate label. I will also integrate the metrics

maschad:
Looks like @dapplion had raised a similar issue https://github.com/ChainSafe/js-libp2p-gossipsub/issues/272 

He also suggested that we could add an array of topics to the signature, what do you think @tuyennhv ?

src/metrics.ts
```diff
@@ -355,6 +355,12 @@ export function getMetrics(
       labelNames: ['topic']
     }),
 
+    duplicateMsgIgnored: register.gauge<{ topic: TopicLabel }>({
+      name: 'gossisub_duplicate_msg_ignored_total',
+      help: 'Total count of duplicate message ignored by topic',
```
tuyennhv, line 360: Need to also mention that this is about the `publish` api, change the name accordingly

src/metrics.ts
```diff
@@ -629,6 +635,11 @@ export function getMetrics(
       }
     },
 
+    onIgnoreDuplicateMsg(topicStr: TopicStr): void {
```
tuyennhv, line 638:
```suggestion
    onIgnorePublishedDuplicateMsg(topicStr: TopicStr): void {
```

tuyennhv:
> Looks like @dapplion had raised a similar issue #272
> 
> He also suggested that we could add an array of topics to the signature, what do you think @tuyennhv ?

I'm not clear about the context in that issue, if we publish same message to multiple topics, we also need to modify the `fastMsgIdFn` for the received side.

@dapplion could you give your ideas on should we implement that in this PR too?

maschad:
Thanks for reviewing @tuyennhv I have made the requested updates. I think we could probably address https://github.com/ChainSafe/js-libp2p-gossipsub/issues/272 in a separate PR since the issue was raised in a different context.

dapplion (dismissed):

twoeths (dismissed):

src/index.ts
```diff
@@ -109,6 +109,8 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
   asyncValidation: boolean
   /** Do not throw `InsufficientPeers` error if publishing to zero peers */
   allowPublishToZeroPeers: boolean
+  /** Do not throw `PublishError.Duplicate` if publishing duplicate messages */
+  ignoreDuplicateMessages: boolean
```
wemeetagain, line 113:
```suggestion
  allowPublishDuplicateMessages: boolean
```

src/types.ts
```diff
@@ -74,6 +74,7 @@ export enum SignaturePolicy {
 
 export type PublishOpts = {
   allowPublishToZeroPeers?: boolean
+  ignoreDuplicateMessages?: boolean
```
wemeetagain, line 77:
```suggestion
allowPublishDuplicateMessages?: boolean
```

src/index.ts
```diff
@@ -2013,9 +2015,16 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     const msgId = await this.msgIdFn(msg)
     const msgIdStr = this.msgIdToStrFn(msgId)
 
+    // Current publish opt takes precedence global opts, while preserving false value
+    const ignoreDuplicateMessages = opts?.ignoreDuplicateMessages ?? this.opts.ignoreDuplicateMessages
```
wemeetagain, line 2019:
```suggestion
    const allowPublishDuplicateMessages = opts?.allowPublishDuplicateMessages ?? this.opts.allowPublishDuplicateMessages
```

src/index.ts
```diff
@@ -2013,9 +2015,16 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
...
+
     if (this.seenCache.has(msgIdStr)) {
       // This message has already been seen. We don't re-publish messages that have already
       // been published on the network.
+      if (ignoreDuplicateMessages) {
+        this.metrics?.onIgnorePublishedDuplicateMsg(topic)
```
wemeetagain, line 2025:
```suggestion
      if (allowPublishDuplicateMessages) {
        this.metrics?.onPublishDuplicateMsg(topic)
```

src/metrics.ts
```diff
@@ -629,6 +635,11 @@ export function getMetrics(
       }
     },
 
+    onIgnorePublishedDuplicateMsg(topicStr: TopicStr): void {
```
wemeetagain, line 638:
```suggestion
    onPublishDuplicateMsg(topicStr: TopicStr): void {
```

wemeetagain (changes requested):

src/types.ts
```diff
@@ -74,6 +74,7 @@ export enum SignaturePolicy {
 
 export type PublishOpts = {
   allowPublishToZeroPeers?: boolean
+allowPublishDuplicateMessages?: boolean
```
wemeetagain, line 77:
```suggestion
  allowPublishDuplicateMessages?: boolean
```

wemeetagain:
> - Added a new option to ignore duplicate messages
> - Updated the metrics for ignoring duplicates

This isn't right. -1 for the AI
Messages aren't being ignored. The metric isn't tracking ignored duplicates.
This feature allows us to publish duplicate messages. The metric tracks the number of published duplicates.

src/index.ts
```diff
@@ -2013,9 +2015,16 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
...
 
+    // Current publish opt takes precedence global opts, while preserving false value
+    const allowPublishDuplicateMessages = opts?.allowPublishDuplicateMessages ?? this.opts.allowPublishDuplicateMessages
+
     if (this.seenCache.has(msgIdStr)) {
       // This message has already been seen. We don't re-publish messages that have already
       // been published on the network.
+      if (allowPublishDuplicateMessages) {
+        this.metrics?.onPublishDuplicateMsg(topic)
+        return { recipients: [] }
```
wemeetagain, line 2026: @tuyennhv correct me if i'm wrong, this isn't what we want?
I thought we want to actually follow the same flow and actually send messages to peers?

wemeetagain (changes requested):

src/index.ts
tuyennhv: in this case some other nodes already published this same message and it's around in the topic peers already, I found it's not helpful to send to topic peers again - some nodes already saw that message and sent it to us, they may already had it in their seen cache too.

In lodestar's scenario, builder published the block, then the node receives the message and try to publish again and it throws error. The published block was spread to the network successfully, only our api throws error.

maschad:
> > * Added a new option to ignore duplicate messages
> > * Updated the metrics for ignoring duplicates
> 
> This isn't right. -1 for the AI Messages aren't being ignored. The metric isn't tracking ignored duplicates. ~This feature allows us to publish duplicate messages. The metric tracks the number of published duplicates.~ Edit: This feature allows us to avoid throwing an error when publishing a duplicate message. The metric tracks the number of duplicate message publish attempts.

Technically it was correct because we where returning early / an empty message which was the original intention in #402 i.e. _option to publish api to return early and ignore the error_ 

As @tuyennhv [mentioned](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/404#discussion_r1111495638) it's not helpful to republish to those peers again since some nodes already would've seen that message, thus they may have already had it in their seen cache too. 

So with that in mind I posit that `ignoreDuplicateMessages` may be a more appropriate name as opposed to `allowPublishDuplicateMessages`

src/index.ts
maschad: Disagree with this for reasons mentioned here https://github.com/ChainSafe/js-libp2p-gossipsub/pull/404#issuecomment-1437401486

wemeetagain:
Reason I'm not a fan of "ignoreDuplicateMessages" naming is that its too general and the terms feel overloaded.
We "ignore" messages in gossip validation, though this feature has nothing to do with this kind of ignoring. We also have two message flows (publish and forward) that handle duplicate messages, so "ignoreDuplicateMessages" sounds like a blanket statement about how we handle duplicate messages.

"allowPublishDuplicateMessage" follows the naming of the other "suppress a publish error" option - "allowPublishToZeroPeers" - generally "allowPublishErrorname".
Do agree there's some confusion in this naming as well tho.

Imo the core feature here is suppressing / avoiding a publish error. Probably there's some other name that can clearly articulate this.

maschad:
Understood. I think `ignoreDuplicatePublishError` captures the essence of conveying that we are trying to avoid a publish error then.

wemeetagain (approved):

PR #410, opened 2023-02-25
Add information about `doPX`

### LINKED ISSUE #409: Newly connecting nodes don't interconnect
- I have 1 bootstrap node and then I launch 3 newly connecting nodes that are all joining via the bootstrap node.
- I subscribe all nodes to a topic name
- When I publish messages under the topic, all nodes receive it too
- However, when I make the bootstrap node go down, sadly now all other nodes have lost connectivity as they were seemingly arranged as a hub and spokes
- My expectation would have been that connectivity among nodes is quickly established. Am I configuring something falsely?

TimDaub:
- Fixes #409

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Options is an optional object with the following key-value pairs:
 - **`gossipIncoming`**: boolean identifying if incoming messages on a subscribed topic should be automatically gossiped (defaults to **true**).
 - **`fallbackToFloodsub`**: boolean identifying whether the node should fallback to the floodsub protocol, if another connecting peer does not support gossipsub (defaults to **true**).
 - **`floodPublish`**: boolean identifying if self-published messages should be sent to all peers, (defaults to **true**).
-- **`doPX`**: boolean identifying whether PX is enabled; this should be enabled in bootstrappers and other well connected/trusted nodes (defaults to **false**).
+- **`doPX`**: boolean identifying whether PX is enabled; this should be enabled in bootstrappers and other well connected/trusted nodes (defaults to **false**). Requires [js-libp2p-pubsub-peer-discovery](https://github.com/libp2p/js-libp2p-pubsub-peer-discovery).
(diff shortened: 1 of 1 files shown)


maschad (changes requested):
We actually mentioned [in the Libp2p discovery mechanism docs](https://github.com/libp2p/js-libp2p/blob/master/examples/discovery-mechanisms/README.md?plain=1#L165-L208) where this is required - I apologize that the docs are bit dated 😓 . 

But if you are using a bootstrap node with `js-libp2p-gossipsub`, you do not necessarily need to use a separate peer discovery mechanism, since the bootstrap node can be used to help your node discover and connect to other peers in the network. The `doPx` mechanism is to optimize the flow of information in the GossipSub network, by allowing peers to make informed decisions about which messages to forward and which peers to forward them to, so I wouldn't say this is required.

wemeetagain (changes requested):
I don't think this is accurate. I do think there's a way to make the docs more helpful, but I'm not sure what exactly to say.

The "PX" feature doesn't enable full connectivity by itself and doesn't necessarily need pubsub-peer-discovery to work. It's just an optional mechanism in gossipsub to have well-connected nodes help poorly-connected nodes to maintain a healthy mesh (by sending them some peers to connect to when they get pruned).

The problem comes because px messages can optionally not include multiaddrs of possible mesh peers and only include peer ids of possible mesh peers (for backwards compatibility sake, since the "signed peer record" spec wasn't yet rolled out when gossipsub v1.1 was released). In that case, when receiving px messages, you'd need a peer discovery configured to hopefully link those possible mesh peers to multiaddrs to dial.

A few points though:
- It's the receiver of the px messages (not necessarily the one who set `doPx = true` that may need peer discovery
- The peer discovery that can or should be used isn't always pubsub-peer-discovery, it depends on the application
- iirc js-libp2p and go-libp2p's implementations of the identify protocol should include signed peer records, so if that's enabled by all peers, then you shouldn't need additional peer discovery to dial possible mesh peers

wemeetagain:
closing this, see the above comment. maybe some libp2p-wide documentation on PX would be helpful to link here.

Issue #413, opened 2023-03-07
Missed Iwant messages from Lodestar to Nimbus


tuyennhv:
There are some `Iwant` messages from Lodestar to Nimbus

```
46955563:Mar-07 05:57:11.503[network]         debug: 2023-03-07T05:57:11.503Z libp2p:gossipsub peer %s didn't follow up in %s IWANT requests; adding penalty 16Uiu2HAmThKa44DYZKvYi52z5gKyAkdLydMrEewrYQHGn9Ez64uS b8cb798a1ec628ef49a3765f755e1950b2d7649c
47035690:Mar-07 05:57:33.767[network]         debug: 2023-03-07T05:57:33.767Z libp2p:gossipsub peer %s didn't follow up in %s IWANT requests; adding penalty 16Uiu2HAmThKa44DYZKvYi52z5gKyAkdLydMrEewrYQHGn9Ez64uS e91fea08007750e2db5db957117e66d53a28b752
47077831:Mar-07 05:57:47.203[network]         debug: 2023-03-07T05:57:47.203Z libp2p:gossipsub peer %s didn't follow up in %s IWANT requests; adding penalty 16Uiu2HAmThKa44DYZKvYi52z5gKyAkdLydMrEewrYQHGn9Ez64uS e7549a3ce238c44d219438b5f7e5f419329959e0
```

however Nimbus does not see these message ids, both side use "hex" as message id string format

tuyennhv:
caught 2 other missed IWANT messages with the same message id (to 2 different nodes), Nimbus confirmed they didn't see this message id at the time

```
grep -e "0xff2cb2265c354d00e4c5fdacd7209d6bc1c65694" -rn beacon-2023-03-08.log
240978708:Mar-08 23:11:36.481[network]         debug: 2023-03-08T23:11:36.481Z libp2p:gossipsub IHAVE: Asking from peers 16Uiu2HAkuzih5wCzH4CxKvnZrekrnjeaz48iEgLb5axDxUbfNj61 message id 0xff2cb2265c354d00e4c5fdacd7209d6bc1c65694
240978826:Mar-08 23:11:36.614[network]         debug: 2023-03-08T23:11:36.614Z libp2p:gossipsub IHAVE: Asking from peers 16Uiu2HAmThKa44DYZKvYi52z5gKyAkdLydMrEewrYQHGn9Ez64uS message id 0xff2cb2265c354d00e4c5fdacd7209d6bc1c65694
[...]

PR #415, opened 2023-03-16
fix: track promise only after a successful sendRpc()

tuyennhv:
**Motivation**
- In rare case, a `sendRpc()` call may be failed or got delayed, we should only track promise after a successful `sendRpc`

part of #413

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1244,7 +1244,11 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
       return
     }
 
-    this.sendRpc(id, { messages: ihave, control: { iwant, prune } })
+    const sent = this.sendRpc(id, { messages: ihave, control: { iwant, prune } })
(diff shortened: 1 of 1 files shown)


tuyennhv:
after 3.5h, there are more than 2k untracked promise due to unsuccessful `sendRpc()` call in a mainnet node with only Nimbus peers

Screenshot 2023-03-16 at 16 39 37

src/index.ts
```diff
@@ -1354,7 +1362,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     iwantList = iwantList.slice(0, iask)
     this.iasked.set(id, iasked + iask)
 
-    this.gossipTracer.addPromise(id, iwantList)
+    // do not add promise here until a successful sendRpc()
```
wemeetagain, line 1365:
```suggestion
    // do not add gossipTracer promise here until a successful sendRpc()
```

wemeetagain (approved):

PR #416, opened 2023-03-23
feat: add onGossipsubMessage

tuyennhv:
**Motivation**
- Right now we use event emitter to emit incoming messages. For an application with busy network i/o, we don't want to keep the event loop busy
- See https://itnext.io/high-performance-node-js-concurrency-with-events-on-hexomter-da4472ae3061

**Description**
- Add `onGossipsubMessage` to emit incoming messages. This is backward compatible while lodestar can simply overwrite `onGossipsubMessage()` without keeping the event loop busy


tuyennhv:
`emit()` does not take time, closing the PR

PR #418, opened 2023-03-24
fix: Potential fix for interop with rust subscriptions

maschad:


src/index.ts
```diff
@@ -2066,6 +2066,8 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
       }
       throw Error('PublishError.Duplicate')
     }
+    // For testing purposes only, see https://filecoinproject.slack.com/archives/C04UFPHSM7S/p1679682078799899?thread_ts=1679674509.496669&cid=C04UFPHSM7S
```
dapplion, line 2069: This conversation can only be viewed if you sign up. Can you add more detail of what's going on?

maschad: The discussion is in a public channel #libp2p-implementers on Filecoin slack (which is bridged to IPFS Discord/Matrix).

The problem [has been resolved now](https://github.com/libp2p/universal-connectivity/pull/3#issuecomment-1485691089) so this is no longer relevant.  Essentially in debugging I had found that there was an issue of `rust-libp2p` sending subscription messages to `js-libp2p-gossipsub` after GRAFT messages over webRTC

PR #432, opened 2023-06-12
fix!: reportMessageValidationResult to accept peer id string

tuyennhv:
**Motivation**
- In lodestar, passing peer id through thread boundary has to be in string, it has to convert back to a `PeerId` instance at lodestar side and `reportMessageValidationResult` to call `toString()` again, it causes lodestar network thread to waste 2.8% of cpu time

Screenshot 2023-06-12 at 10 42 28

**Description**
- reportMessageValidationResult to accept peer id string instead, it's a breaking change

part of https://github.com/ChainSafe/lodestar/issues/5604

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2139,7 +2139,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
    *
    * This should only be called once per message.
    */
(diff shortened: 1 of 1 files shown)


dapplion (approved):
LGTM!

PR #434, opened 2023-06-12
Use PushableAbortable

dapplion:
**Motivation**

In high traffic apps like a beacon node, abortableSource takes +1% of total CPU time.

**Description**

This PR removes the need for one loop of abortableSource in each outbound packet. Uses a simplified implementation of pushable with abort capacity using Lodestar's LinkedList https://github.com/ChainSafe/lodestar/blob/09ecf9957c33c3bdf79645dbd27979bdb0514c56/packages/beacon-node/src/util/array.ts#L32 and AsyncIterator from https://github.com/ChainSafe/lodestar/blob/09ecf9957c33c3bdf79645dbd27979bdb0514c56/packages/beacon-node/src/util/asyncIterableToEvents.ts#L53

### DIFF
--- a/src/stream.ts
+++ b/src/stream.ts
@@ -1,9 +1,9 @@
 import { Stream } from '@libp2p/interface-connection'
 import { abortableSource } from 'abortable-iterator'
 import { pipe } from 'it-pipe'
-import { pushable, Pushable } from 'it-pushable'
 import { encode, decode } from 'it-length-prefixed'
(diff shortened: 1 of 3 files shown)


wemeetagain:
We might wait for https://github.com/libp2p/js-libp2p/issues/1793#issue-1741818809 (check last few paragraphs of the comment) to be implemented which will give us abortable sources at a lower level.

wemeetagain:
closing in favor of #488

PR #444, opened 2023-06-28
fix: check backoff when join

tuyennhv:
**Motivation**
- To conform to the gossipsub spec v1.1

**Description**
- add some slack time to the backoff expiration https://github.com/libp2p/specs/pull/289
- do not send GRAFT to any backoff peer, otherwise we'll immediately receive PRUNE and we'll add more backoff time to that peer


src/index.ts
```diff
@@ -1643,9 +1644,12 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     }
 
     const now = Date.now()
+    const slackDuration = BACKOFF_SLACK * this.opts.heartbeatInterval
     this.backoff.forEach((backoff, topic) => {
       backoff.forEach((expire, id) => {
-        if (expire < now) {
+        // add some slack time to the expiration
+        // https://github.com/libp2p/specs/pull/289
+        if (expire + slackDuration < now) {
```
dapplion, line 1652:
```suggestion
        // add some slack time to the expiration, see https://github.com/libp2p/specs/pull/289
        if (expire + BACKOFF_SLACK * this.opts.heartbeatInterval < now) {
```

src/index.ts
```diff
@@ -1790,6 +1794,7 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     this.metrics?.onJoin(topic)
 
     const toAdd = new Set<PeerIdStr>()
+    const backoff = this.backoff.get(topic) || new Map()
```
dapplion, line 1797: Do not allocate a new map if not necessary, handle the undefined value case below

wemeetagain (approved):

PR #445, opened 2023-06-28
chore: track backoff time for connected peers

### LINKED ISSUE #442: Track backoff times in connected peers

tuyennhv:
**Motivation**
- To track backoff time for connected peers

**Description**
- Add respective metric
Closes #442

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2873,10 +2873,17 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
     metrics.cacheSize.set({ cache: 'control' }, this.control.size)
     metrics.cacheSize.set({ cache: 'peerhave' }, this.peerhave.size)
     metrics.cacheSize.set({ cache: 'outbound' }, this.outbound.size)
+
     // 2D nested data structure
     let backoffSize = 0
+    const now = Date.now()
     for (const backoff of this.backoff.values()) {
       backoffSize += backoff.size
+      for (const [peer, expiredMs] of backoff.entries()) {
+        if (this.peers.has(peer)) {
+          metrics.connectedPeersBackoffSec.observe((expiredMs - now) / 1000)
+        }
+      }
     }
     metrics.cacheSize.set({ cache: 'backoff' }, backoffSize)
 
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -2873,10 +2873,17 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
...
     metrics.cacheSize.set({ cache: 'outbound' }, this.outbound.size)
+
     // 2D nested data structure
     let backoffSize = 0
+    const now = Date.now()
     for (const backoff of this.backoff.values()) {
       backoffSize += backoff.size
+      for (const [peer, expiredMs] of backoff.entries()) {
+        if (this.peers.has(peer)) {
+          metrics.connectedPeersBackoffSec.observe((expiredMs - now) / 1000)
```
dapplion, line 2884: Reset the histogram values above to have a read of current status. Similar to Lodestar's gossipsub metrics. Also record only positive numbers or histograms render poorly `Math.max(0, expiredMs - now) / 1000`

src/metrics.ts
```diff
@@ -554,6 +554,12 @@ export function getMetrics(
       name: 'gossip_iwant_promise_untracked',
       help: 'Total count of untracked IWANT promise'
     }),
+    /** Backoff time */
+    connectedPeersBackoffSec: register.histogram({
+      name: 'gossipsub_connected_peers_backoff_seconds',
+      help: 'Backoff time in seconds',
+      buckets: [1, 2, 4, 8, 16, 32, 64, 128, 256]
```
dapplion, line 561: Values could align more with spec parameters such that
```
[1,2,4,10,20,60,120]
```

also do add a comment like:

```
Using 1 seconds as minimum as that's close to the heartbeat duration, no need for more resolution.
As per spec, backoff times are 10 seconds for UnsubscribeBackoff and 60 seconds for PruneBackoff.
Higher values of 60 seconds should not occur, but we add 120 seconds just in case
https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#overview-of-new-parameters
```

wemeetagain (approved):

PR #447, opened 2023-06-29
feat: add UnsubscribeBackoff param

### LINKED ISSUE #249: New parameter: pruneBackoff and unsubcribeBackoff
**Motivation**
+ Right now we have backoff time as a constant of 1 minute 
```typescript
export const GossipsubPruneBackoff = minute
```
+ Lodestar wants to specifies its own `backoff` time when it unsubscribes from a topic
+ Refer to nim's implementation https://github.com/status-im/nim-libp2p/pull/665

**Description**
+ Support `unsubscribeBackoff` param when we unsubscribe from a topic, default to some seconds
+ Support `pruneBackoff` param for other scenarios, default to 1 minute

tuyennhv:
**Motivation**
- We want to specify lower backoff time if unsubscribing from a topic as we may resubscribe from that same topic soon
- To follow v1.1 spec of gossipsub https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md

**Description**
- In `makePrune()`, add a new parameter onUnsubscribe. If it's true specify the backoff time as UnsubscribeBackoff param, update all consumers accordingly
- Also found that we missed adding backoff in this function, double checked with rust-libp2p https://github.com/libp2p/rust-libp2p/blob/f335697573e2688390deeb9708abaeec9f0b9d9f/protocols/gossipsub/src/behaviour.rs#L1116

Closes #249

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -133,6 +133,7 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
   /** override constants for fine tuning */
   prunePeers?: number
   pruneBackoff?: number
+  unsubcribeBackoff?: number
   graftFloodThreshold?: number
   opportunisticGraftPeers?: number
   opportunisticGraftTicks?: number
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -2408,16 +2423,19 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
       }
     }
     // backoff is measured in seconds
-    // GossipsubPruneBackoff is measured in milliseconds
+    // GossipsubPruneBackoff and GossipsubUnsubscribeBackoff are measured in milliseconds
     // The protobuf has it as a uint64
-    const backoff = this.opts.pruneBackoff / 1000
+    const backoffMs = onUnsubscribe ? this.opts.unsubcribeBackoff : this.opts.pruneBackoff
```
wemeetagain, line 2428:
```suggestion
    const backoffMs = onUnsubscribe ? this.opts.unsubcribeBackoff : this.opts.pruneBackoff
    const backoff = backoffMs / 1000
```

wemeetagain (approved):

PR #451, opened 2023-07-16
feat: track time to publish a message

tuyennhv:
**Motivation**
- I'd like to know how much time to publish a message at gossipsub side

**Description**
- Add `msgPublishMs` metric and track it

related to #344

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -338,6 +338,13 @@ export function getMetrics(
       help: 'Total count of msg publish data.length bytes',
       labelNames: ['topic']
     }),
+    /** Total time in millisecond to publish a message */
+    msgPublishMs: register.histogram<{ topic: TopicLabel }>({
+      name: 'gossipsub_msg_publish_ms',
+      help: 'Total time in millisecond to publish a message',
+      buckets: [5, 10, 100, 500, 1000],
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

wemeetagain (approved):

tuyennhv:
we can track metric in `ns`, however if it takes <1ms consistently to publish data then I suppose we don't have a big performance issue here

src/metrics.ts
```diff
@@ -338,6 +338,13 @@ export function getMetrics(
       help: 'Total count of msg publish data.length bytes',
       labelNames: ['topic']
     }),
+    /** Total time in millisecond to publish a message */
+    msgPublishMs: register.histogram<{ topic: TopicLabel }>({
+      name: 'gossipsub_msg_publish_ms',
+      help: 'Total time in millisecond to publish a message',
```
dapplion, line 344: Prometheus standards strongly suggest to use standard metrics, in the case of time seconds: https://prometheus.io/docs/practices/naming/#base-units

tuyennhv: thanks @dapplion I created https://github.com/ChainSafe/js-libp2p-gossipsub/pull/457/files

PR #460, opened 2023-08-07
feat: unbundle fixed-label metrics

tuyennhv:
**Motivation**
- Not able to answer this question: what's the % of IGNORED/REJECTED message validation results for a specific topic
See https://github.com/ChainSafe/lodestar/issues/5853

**Description**
- Revise the work in https://github.com/ChainSafe/js-libp2p-gossipsub/pull/433/files
- If the label is gossipsub/fixed, we should unbundle to multiple metrics to answer the above question

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -1,4 +1,4 @@
-import type { TopicValidatorResult } from '@libp2p/interface/pubsub'
+import { TopicValidatorResult } from '@libp2p/interface/pubsub'
 import type { IRPC } from './message/rpc.js'
 import type { PeerScoreThresholds } from './score/peer-score-thresholds.js'
 import {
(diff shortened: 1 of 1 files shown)


src/metrics.ts
```diff
@@ -626,8 +700,29 @@ export function getMetrics(
...
+          break
+        case InclusionReason.NotEnough:
+          this.meshPeerInclusionEventsNotEnough.inc({ topic }, count)
+          break
+        case InclusionReason.Opportunistic:
+          this.meshPeerInclusionEventsOpportunistic.inc({ topic }, count)
+          break
+        default:
+          this.meshPeerInclusionEventsUnknown.inc({ topic }, count)
+          break
```
dapplion, line 724: Is it really necessary to have so much granularity here?

tuyennhv: @dapplion  without it we don't know how mesh peers are added for `beacon_attestation` topic like before

maybe it's not important for this specific metric, but the same granularity for `ChurnReason` metrics below are quite important when investigating low mesh peer issue

so it's more code here but beneficial for the applications using gossipsub

dapplion: Okk sounds good

dapplion (approved):

wemeetagain (approved):

PR #464, opened 2023-08-15
fix: revise onPrevalidationResult metrics

tuyennhv:
**Motivation**
- Found duplicate metrics after merging #460
- There are validation at gossipsub side, and validation at application side that cause the duplicate metric names

**Description**
- Name the validation in gossipsub "preValidationResult" to distinguish to the validation result coming from application
- Also fix `meshPeerChurnEventsUnknown` metric

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -261,7 +261,7 @@ export function getMetrics(
       labelNames: ['topic']
     }),
     meshPeerChurnEventsUnknown: register.gauge<{ topic: TopicLabel }>({
-      name: 'gossipsub_peer_churn_events_excess_total',
+      name: 'gossipsub_peer_churn_events_unknown_total',
       help: 'Number of times we remove peers in a topic mesh for unknown reasons',
       labelNames: ['topic']
     }),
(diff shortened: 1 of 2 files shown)


dapplion (approved):

src/metrics.ts
```diff
@@ -443,23 +443,23 @@ export function getMetrics(
...
-      name: 'gossipsub_msg_received_duplicate_total',
+    prevalidationDuplicateTotal: register.gauge<{ topic: TopicLabel }>({
+      name: 'gossipsub_pre_validation_duplicate_total',
       help: 'Total count of duplicate messages received',
       labelNames: ['topic']
     }),
-    msgReceivedUnknownStatusTotal: register.gauge<{ topic: TopicLabel }>({
-      name: 'gossipsub_msg_received_unknown_status_total',
+    prevalidationUnknownTotal: register.gauge<{ topic: TopicLabel }>({
+      name: 'gossipsub_pre_validation_unknown_status_total',
```
dapplion, line 462: hmm is the duplicate metric on lodestar or here? we should try to not break metrics on upstream libraries.

dapplion (changes requested):

src/metrics.ts
tuyennhv: this metric was introduced in the previous PR https://github.com/ChainSafe/js-libp2p-gossipsub/pull/460/files#diff-a210f4dc1ff876b56a2a9db45167223854fcbf9b9dc6bd0b4e92433582a3d6f8R462

no one uses that metric for now @dapplion

PR #466, opened 2023-08-30
chore: track total metric for mesh inclusion/churn events

tuyennhv:
**Motivation**
We removed `gossipsub_mesh_peer_inclusion_events_total` and `gossipsub_peer_churn_events_total` in #460 but lodestar still needs it, see https://github.com/ChainSafe/lodestar/issues/5919

**Description**
Add back these metrics

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -213,6 +213,11 @@ export function getMetrics(
       help: 'Number of times we include peers in a topic mesh for random reasons',
       labelNames: ['topic']
     }),
+    meshPeerInclusionEventsTotal: register.gauge<{ topic: TopicLabel }>({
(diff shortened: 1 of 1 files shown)


matthewkeil (approved):
LGTM 🎸

src/metrics.ts
```diff
@@ -213,6 +213,11 @@ export function getMetrics(
       help: 'Number of times we include peers in a topic mesh for random reasons',
       labelNames: ['topic']
     }),
+    meshPeerInclusionEventsTotal: register.gauge<{ topic: TopicLabel }>({
```
wemeetagain, line 216: maybe we can rename these metrics. I think "inclusion/churn" is a bit more confusing than "add/remove"

PR #470, opened 2023-10-31
fix: use typed event emitter class

achingbrain:
The typed `EventEmitter` class from `@libp2p/interface/events` has been renamed `TypedEventEmitter` along with a `TypedEventTarget` interface.

The old name is still exported so it's a backwards compatible change but it will be removed in a future version of `@libp2p/interface` so this change future-proofs `@chainsafe/libp2p-gossipsub` against that upcoming release.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-import { CustomEvent, EventEmitter } from '@libp2p/interface/events'
+import { CustomEvent, TypedEventEmitter } from '@libp2p/interface/events'
 import { StrictSign, StrictNoSign, TopicValidatorResult } from '@libp2p/interface/pubsub'
 import { type Logger, logger } from '@libp2p/logger'
 import { peerIdFromBytes, peerIdFromString } from '@libp2p/peer-id'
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -1024,11 +1024,9 @@ export class GossipSub extends EventEmitter<GossipsubEvents> implements PubSub<G
         }
       })
 
-      this.dispatchEvent(
-        new CustomEvent<SubscriptionChangeData>('subscription-change', {
-          detail: { peerId: from, subscriptions }
-        })
-      )
+      this.safeDispatchEvent<SubscriptionChangeData>('subscription-change', {
```
achingbrain, line 1027: This style of invocation type-checks the `detail` field.

achingbrain:
Looks like the custom linting rules in this repo are broken.  [Upgrading @typescript-eslint/parser may help](https://stackoverflow.com/questions/76769634/why-am-i-getting-a-deprecationwarning-originalkeywordkind-deprecated-error) - or since aegir is being used to build this project, perhaps delete all the custom eslint stuff and just use aegir's linting?

wemeetagain (approved):

PR #479, opened 2024-01-03
feat: type safe metric labels

nflaig:
See https://github.com/ChainSafe/lodestar/pull/6201 for motivation / rationale

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -19,58 +19,45 @@ export enum MessageSource {
   publish = 'publish'
 }
 
-type LabelsGeneric = Record<string, string | undefined>
+type NoLabels = Record<string, never>
+type LabelsGeneric = Record<string, string | number>
+type LabelKeys<Labels extends LabelsGeneric> = Extract<keyof Labels, string>
 interface CollectFn<Labels extends LabelsGeneric> { (metric: Gauge<Labels>): void }
 
-export interface Gauge<Labels extends LabelsGeneric = never> {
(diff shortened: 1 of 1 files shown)


src/metrics.ts
```diff
@@ -328,10 +315,9 @@ export function getMetrics (
       labelNames: ['hit']
     }),
 
-    asyncValidationDelayFromFirstSeenSec: register.histogram<{ topic: TopicLabel }>({
+    asyncValidationDelayFromFirstSeenSec: register.histogram({
```
nflaig, line 318: Label value is not set for this metric
https://github.com/ChainSafe/js-libp2p-gossipsub/blob/eb53ea96a765a4aeed19b4feee273126a8286616/src/metrics.ts#L793

Topic might not be available when setting the metric hence I updated the type instead of updating the code and setting the topic.

src/metrics.ts
```diff
@@ -883,9 +869,9 @@ export function getMetrics (
     },
 
     onDuplicateMsgDelivery (topicStr: TopicStr, deliveryDelayMs: number, isLateDelivery: boolean): void {
-      this.duplicateMsgDeliveryDelay.observe(deliveryDelayMs / 1000)
+      const topic = this.toTopic(topicStr)
+      this.duplicateMsgDeliveryDelay.observe({ topic }, deliveryDelayMs / 1000)
```
nflaig, line 873: Label was defined in type but not set. I'd assume this was just an oversight as topic is always available here.

dapplion (approved):

PR #480, opened 2024-01-04
feat: batch publish

### LINKED ISSUE #344: Publish in batch

tuyennhv:
**Description**
- Add batch option to publish() api and GossipSub option
- When it's true, encode messages once and send to all peer, ignore cached control messages

cc @wemeetagain 

Closes #344

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -73,7 +73,8 @@ export enum SignaturePolicy {
 export interface PublishOpts {
   allowPublishToZeroPeers?: boolean
   ignoreDuplicatePublishError?: boolean
-  batch?: boolean
+  /** serialize message once and send to all peers without control messages */
+  batchPublish?: boolean
 }
 
 export enum PublishConfigType {
--- a/src/types.ts
+++ b/src/types.ts
@@ -73,6 +73,7 @@ export enum SignaturePolicy {
 export interface PublishOpts {
   allowPublishToZeroPeers?: boolean
   ignoreDuplicatePublishError?: boolean
+  batch?: boolean
 }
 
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -73,6 +73,7 @@ export enum SignaturePolicy {
 export interface PublishOpts {
   allowPublishToZeroPeers?: boolean
   ignoreDuplicatePublishError?: boolean
+  batch?: boolean
```
wemeetagain, line 76:
```suggestion
  /** serialize message once and send to all peers without control messages */
  batchPublish?: boolean
```

dapplion (approved):

wemeetagain (approved):

PR #484, opened 2024-01-18
fix: write peer stream messages atomically

achingbrain:
The sink passed to a libp2p stream will have chunks pulled from it when the previous chunk has read by the underlying transport. This allows the transport to apply backpressure all the way up to the application level without libp2p needing to buffer application data which can be very memory-intensive.

The batch-message writing in #344 introduces a [second pushable queue](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/master/src/stream.ts#L30-L38) to peer streams - messages from both queues are written to the underlying stream in parallel, removing any guarantees of order or consistency.

Notably if we are writing a `Uint8ArrayList` made of multiple `Uint8Array`s, the underlying transport will pull the arrays one at a time to avoid a memory copy, if the stream is written to in parallel all bets are off in regard to the order in which the arrays will be sent over the wire.

This manifests itself as the remote failing to read a message because the message length is too long - it's read some random garbage from the middle of another message as a varint representing the length of the current message.

The change here is to combine the prefixed and non-prefixed pushable streams with `it-merge`, allowing the messages to be pulled in a predictable way and preserving the ability of the underlying transport to apply backpressure.

Also updates all deps to reduce duplication in web bundles.

Notably `multiformats@13` is the TypeScript port and `@libp2p/crypto@4` has dropped the `node-forge` dependency.

This drops the bundle size from around `132KB`  to `85KB`

Refs:
- https://github.com/libp2p/js-libp2p/issues/1793
- https://github.com/achingbrain/it/pull/103


wemeetagain (approved):

wemeetagain:
The e2e tests are consistently failing :thinking:

PR #487, opened 2024-02-02
fix!: make PublishError.InsufficientPeers more self-explanatory

### LINKED ISSUE #472: Make `PublishError.InsufficientPeers` a bit more self-explanatory?
A recurring question that comes up is:

> I get `PublishError.InsufficientPeers` but I have plenty of peers, what gives?

The error does not mean you don't have peers, it means you don't have any peers listening on the pubsub topic you are trying to publish on.

If this was changed to `PublishError.InsufficientTopicPeers` or even something like `PublishError.InsufficientPeersSubscribedToTopic` it might be a bit clearer for people?

achingbrain:
People continually mistake the `InsufficientPeers` error for something that can be "worked around" by setting `allowPublishToZeroPeers` to true, and they then expect other network nodes to still recieve their messages.

This PR renames the option to `allowPublishToZeroTopicPeers` - this makes it clear(er) that it's not that you don't have any peers, it's that no peers you have are listening on the topic.

It also improves the JSDoc comment on the option and changes the `Error` message from `PublishError.InsufficientPeers` to `PublishError.NoPeersSubscribedToTopic` which (I hope) more accurately describes what's wrong.

BREAKING CHANGE: The `allowPublishToZeroPeers` option has been renamed to `allowPublishToZeroTopicPeers`

Fixes #472

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -104,8 +104,15 @@ export interface GossipsubOpts extends GossipsubOptsSpec, PubSubInit {
    * reportMessageValidationResult() after the message is dropped from mcache won't forward the message.
    */
   asyncValidation: boolean
-  /** Do not throw `InsufficientPeers` error if publishing to zero peers */
-  allowPublishToZeroPeers: boolean
+  /**
+   * Do not throw `PublishError.NoPeersSubscribedToTopic` error if there are no
+   * peers listening on the topic.
+   *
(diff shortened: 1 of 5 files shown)


achingbrain:
I'm not sure this is worth a major bump on it's own, maybe you want to batch it up with #468?

PR #488, opened 2024-02-02
fix: remove abortable iterator

achingbrain:
AbortableSource is slow because it races promises against every chunk causing extra async work.

It's only really necessary if we're going to pass the source off to another component.

Here we don't do that so it's simpler to just add a listener for the abort event and close the stream.

### DIFF
--- a/src/stream.ts
+++ b/src/stream.ts
@@ -1,4 +1,3 @@
-import { abortableSource } from 'abortable-iterator'
 import { encode, decode } from 'it-length-prefixed'
 import { pipe } from 'it-pipe'
 import { pushable, type Pushable } from 'it-pushable'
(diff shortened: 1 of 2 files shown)


src/stream.ts
```diff
@@ -59,7 +65,6 @@ export class OutboundStream {
     this.closeController.abort()
     // similar to pushable.end() but clear the internal buffer
     await this.pushable.return()
```
achingbrain, line 67: I don't think this is necessary since the `pushable` is only referenced by this stream class so everything will be garbage collected shortly after the `close` method is called on the `OutboundStream` instance.

```suggestion
```

achingbrain:
#434 can probably be closed if this is merged.

wemeetagain (approved):

PR #493, opened 2024-03-27
feat: publish at least mesh_n peers

tuyennhv:
**Motivation**
- Lodestar may want to set `floodPublish` as false in the future

**Description**
- In that case we want to publish at least `mesh_n` peers

see https://github.com/ChainSafe/lodestar/issues/6596#issuecomment-2021834289

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2037,6 +2037,8 @@ export class GossipSub extends TypedEventEmitter<GossipsubEvents> implements Pub
             tosendCount.mesh++
           })
 
+          // We want to publish to at least `D` peers.
+          // If there are insufficient peers in the mesh, publish to other topic peers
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -2036,6 +2036,18 @@ export class GossipSub extends TypedEventEmitter<GossipsubEvents> implements Pub
             tosend.add(peer)
             tosendCount.mesh++
           })
+
+          if (meshPeers.size < this.opts.D) {
```
wemeetagain, line 2040:
```suggestion
          // We want to publish to at least `D` peers.
          // If there are insufficient peers in the mesh, publish to other topic peers
          if (meshPeers.size < this.opts.D) {
```

wemeetagain (approved):

PR #502, opened 2024-08-17
feat: add `gossipFactor` parameter to `GossipsubOpts`

### LINKED ISSUE #501: `GossipFactor` is not configurable
Based on [the spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#overview-of-new-parameters), the `GossipFactor` is a configurable parameter with a reasonable default of `0.25`. But in the current implementation, it is supposed that it's a constant.  
What is the reason the team didn't include it in the options, like other params such as `FloodPublish` or `PruneBackoff`?

mkermani144:
Closes #501

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -9,7 +9,10 @@ export interface GossipsubOptsSpec {
   Dscore: number
   /** Dout sets the quota for the number of outbound connections to maintain in a topic mesh. */
   Dout: number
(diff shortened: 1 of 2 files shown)


wemeetagain (dismissed):

twoeths (approved):

wemeetagain (approved):
Thanks @mkermani144

PR #508, opened 2024-09-24
feat: add ethereum/beacon-metrics per spec

matthewkeil:
More metrics were added to `ethereum/beacon-metrics` to help track peerDAS.  There were also a couple of adds related to gossipsub so made those changes here.

**NOTE:** There is a breaking change to the metrics.  It is not an API break but the metric names changes such that dashboards will not show up correctly without changing the PromQL 

A few of the metric name changed so a PR to lodestar will need to be made when updating the dependency to modify the dashboards so they reference the updated metric names

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1212,11 +1212,12 @@ export class GossipSub extends TypedEventEmitter<GossipsubEvents> implements Pub
    * May forward to all peers in the mesh.
    */
   private async handleReceivedMessage (from: PeerId, rpcMsg: RPC.Message): Promise<void> {
-    this.metrics?.onMsgRecvPreValidation(rpcMsg.topic)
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -2294,7 +2295,7 @@ export class GossipSub extends TypedEventEmitter<GossipsubEvents> implements Pub
       topic,
       tosendCount,
       tosend.size,
-      rawMsg.data != null ? rawMsg.data.length : 0,
+      RPC.Message.getSize(rawMsg),
```
matthewkeil, line 2298: I noticed this, we were only counting the "data" portion but more than that gets sent over the wire in a `RPC.message`.  I updated with this method but not sure if that is "correct" because I did not find anything in the spec speaking to this.

PR #512, opened 2025-01-11
fix: in order for the isPubSub function to work with gossipsub this is mandatory

sfroment:
Signed-off-by: Sacha Froment <sfroment42@gmail.com>


achingbrain (approved):

achingbrain:
Small nit but it might be worth adding a test that asserts instances of `GosspSub` make `isPubSub` return true?

sfroment:
Will add and won't forget such stuff in the future 🙏

sfroment:
@wemeetagain

PR #521, opened 2025-04-16
fix(deps-dev): Dynamic import node:events

tysonclugg:
Avoids the following error when running `pnpm run test`:
```
test browser
ℹ Browser "chromium" setup complete.
✘ [ERROR] Could not resolve "events"

    dist/test/utils/create-pubsub.js:1:32:
      1 │ import { setMaxListeners } from 'events';
        │                                 ~~~~~~~~
        ╵                                 "./events"

  The package "events" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.
```

Also allows errors not related to failed imports to propagate correctly.

### DIFF
--- a/test/utils/create-pubsub.ts
+++ b/test/utils/create-pubsub.ts
@@ -1,4 +1,3 @@
-import { setMaxListeners } from 'events'
 import { generateKeyPair } from '@libp2p/crypto/keys'
 import { TypedEventEmitter, start } from '@libp2p/interface'
 import { mockRegistrar, mockConnectionManager, mockNetwork } from '@libp2p/interface-compliance-tests/mocks'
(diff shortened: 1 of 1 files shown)


test/utils/create-pubsub.ts
```diff
@@ -1,4 +1,3 @@
-import { setMaxListeners } from 'events'
```
achingbrain, line 1: You should use `import { setMaxListeners } from '@libp2p/interface'` - it is a no-op in a browser, then you don't need to add a dynamic import here.

tysonclugg:
@achingbrain Now importing from `@libp2p/interface` as requested, ready for review.

PR #522, opened 2025-04-16
fix(deps-dev): Add missing transient dev dependencies

tysonclugg:
Avoids issues with `pnpm install` and `pnpm run test` such as this:
```
test browser
ℹ Browser "chromium" setup complete.
✘ [ERROR] Could not resolve "mocha/mocha.js"

    <stdin>:8:18:
      8 │ import mocha from 'mocha/mocha.js'
        ╵                   ~~~~~~~~~~~~~~~~

  You can mark the path "mocha/mocha.js" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle.
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,6 +31,7 @@
         "@libp2p/interface-compliance-tests": "^6.0.0",
         "@libp2p/logger": "^5.0.0",
         "@libp2p/peer-store": "^11.0.0",
+        "@types/mocha": "^10.0.10",
         "@types/node": "^20.11.6",
         "@types/sinon": "^17.0.3",
         "abortable-iterator": "^5.1.0",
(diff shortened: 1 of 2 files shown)


achingbrain:
If these modules are not referenced directly in code, they're likely to break `aegir dep-check`.  At least it would, if it was used here.

Is this a problem that using `pnpm` has introduced?

tysonclugg:
> If these modules are not referenced directly in code, they're likely to break `aegir dep-check`. At least it would, if it was used here.

Ahh, you're right. It seems that a different package is missing the dependency.

> Is this a problem that using `pnpm` has introduced?

The problem becomes visible when using `pnpm`, but ultimately it's still a dependency bug.  But as you've pointed out, it's not a bug with this package directly.

Happy to close, thanks. :-)

PR #530, opened 2025-09-12
fix: fix the nosign policy validation wrong check

GrapeBaBa:
This pull request makes a small fix in the `validateToRawMessage` function to correctly check the presence of the `from` field instead of the `key` field when validating messages with the `StrictNoSign` option. This improves the accuracy of validation error reporting.

### DIFF
--- a/src/utils/buildRawMessage.ts
+++ b/src/utils/buildRawMessage.ts
@@ -91,7 +91,7 @@ export async function validateToRawMessage (
     case StrictNoSign:
       if (msg.signature != null) return { valid: false, error: ValidateError.SignaturePresent }
       if (msg.seqno != null) return { valid: false, error: ValidateError.SeqnoPresent }
-      if (msg.key != null) return { valid: false, error: ValidateError.FromPresent }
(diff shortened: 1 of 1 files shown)


src/utils/buildRawMessage.ts
```diff
@@ -91,7 +91,7 @@ export async function validateToRawMessage (
     case StrictNoSign:
       if (msg.signature != null) return { valid: false, error: ValidateError.SignaturePresent }
       if (msg.seqno != null) return { valid: false, error: ValidateError.SeqnoPresent }
-      if (msg.key != null) return { valid: false, error: ValidateError.FromPresent }
```
wemeetagain, line 94: I suppose we also want to keep the validation that msg.key is not present (with a ValidateError.KeyPresent error)

GrapeBaBa: If there is indeed such a rule, I would like to add it. The original code is consistent with the implementation of rust, and the implementation of rust does not check the key.

wemeetagain (approved):

PR #531, opened 2025-09-17
fix: remove duplicate pubkey conversion

GrapeBaBa:
This pull request makes a small change to the `validateToRawMessage` function in `src/utils/buildRawMessage.ts`, ensuring that the `key` property is always set to the provided `publicKey` rather than using `publicKeyFromProtobuf(msg.key)` again because it had been called in the previous code.


wemeetagain (approved):
Thanks, good catch

PR #543, opened 2026-03-12
fix: prune empty topic entries from topics map

### LINKED ISSUE #542: Memory leak: prune empty topic entries from gossipsub topics map
## Summary

While investigating a network-thread memory leak in Lodestar, I found a secondary leak pattern in `@libp2p/gossipsub`: the internal `topics` map retains empty topic entries indefinitely.

`topics` is declared as:

```ts
private readonly topics = new Map<TopicStr, Set<PeerIdStr>>()
```

Entries are created for every topic string seen in remote subscription updates, but empty entries are never pruned when the last peer unsubscribes or disconnects.

### Entry creation
`handleReceivedSubscription()` creates a map entry on first sight of a topic:

```ts
let topicSet = this.topics.get(topic)
if (topicSet == null) {
  topicSet = new Set()
  this.topics.set(topic, topicSet)
}
```

### Missing deletion on unsubscribe
When a peer unsubscribes, only the peer id is removed from the set:

```ts
topicSet.delete(from.toString())
```

If that was the last peer, the now-empty `Set` remains in `this.topics` forever.

### Missing deletion on disconnect
Similarly, `removePeer()` iterates all topic sets and deletes the peer id:

```ts
for (const peers of this.topics.values()) {
  peers.delete(id)
}
```

Again, empty sets are left behind and the topic key is never removed.

## Why this matters

On Ethereum consensus clients, peers advertise many topic strings over time (fork-specific gossip topics, attestation subnets, etc.). If empty topic entries are retained forever, `this.topics.size` grows monotonically with topic churn.

That has two effects:

1. **Direct retention**: each stale topic entry retains the topic string key + empty `Set`
2. **Secondary allocation pressure**: heartbeat and publish paths iterate topic structures and allocate per-topic helpers / arrays

## Heap evidence from a real Lodestar node

From network-thread heap snapshots on a production-like node:

- post-deploy snapshot @ `2026-03-12T11:50:05Z`: **2,177,153** nodes
- later snapshot @ `2026-03-12T22:00:01Z`: **2,838,135** nodes
- delta over ~10h: **+660,982** nodes

Relevant class growth over that 10h window:

- `Set`: **38,022 → 53,540** (**+15,518**)
- `Object`: **550,382 → 754,118** (**+203,736**)
- `Buffer`: **+1,813**
- `Uint8Array`: **+1,972**
- `ArrayBuffer`: **+1,365**

At the same time, the original req/resp leak suspects were flat or decreasing:

- `Connection`: **-5**
- `Socket`: **-6**
- `MplexStream`: **+4**

We also saw growth in retained topic strings such as:

- `beacon_attestation_35`: **4 → 321**
- `beacon_attestation_57`: **5 → 302**

This points away from stream/socket retention and toward topic bookkeeping.

## Minimal fix

Prune topic entries when the last peer is removed.

### On unsubscribe

```ts
topicSet.delete(from.toString())
if (topicSet.size === 0) {
  this.topics.delete(topic)
}
```

### On peer disconnect

```ts
for (const [topic, peers] of this.topics) {
  peers.delete(id)
  if (peers.size === 0) {
    this.topics.delete(topic)
  }
}
```

## Tests to add

Two regression tests seem sufficient:

1. **remote unsubscribe cleanup**
   - peer A subscribes to topic
   - peer B receives subscription update
   - peer A unsubscribes
   - assert peer B no longer retains `topics.has(topic)`

2. **peer disconnect cleanup**
   - peer A subscribes to topic
   - peer B receives subscription update
   - peer A disconnects
   - assert peer B no longer retains `topics.has(topic)`

## Notes

This is distinct from the req/resp clearable-signal leak we were originally debugging. That original leak appears fixed; this issue showed up only after the first fix reduced the baseline enough to expose the next growth vector.

If helpful, I can turn the minimal fix above into a PR.

lodekeeper:
## Summary

Fix a `topics` map retention bug in gossipsub.

When a remote peer unsubscribed from a topic, or disconnected after being the last subscriber for a topic, the internal `topics: Map<TopicStr, Set<PeerIdStr>>` entry was left behind as an empty `Set`. Over time this let topic strings accumulate indefinitely.

This patch:
- prunes empty `topics` entries on remote unsubscribe
- prunes empty `topics` entries on peer disconnect
- avoids creating empty `topics` entries from unsubscribe-only updates
- adds regression tests for all of the above

Closes #542.

## Details

The minimal fix is intentionally scoped to `topics` only.

I did **not** prune `mesh` entries in the same change because `mesh` is local joined-topic state and deleting empty mesh topics on disconnect would interfere with heartbeat maintenance/regrafting. `fanout` already has TTL cleanup, and `backoff` already self-prunes empty per-topic maps.

## Testing

- `pnpm test -- --target node --grep 'topic entries|unsubscribe-only'`
- `pnpm lint`

## AI disclosure

Initial heap analysis and patch drafting were assisted by AI tooling, then validated with local source inspection and test runs.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -963,8 +963,11 @@ export class GossipSub extends TypedEventEmitter<GossipsubEvents> implements Pub
     this.streamsInbound.delete(id)
 
     // remove peer from topics map
-    for (const peers of this.topics.values()) {
+    for (const [topic, peers] of this.topics) {
       peers.delete(id)
+      if (peers.size === 0) {
+        this.topics.delete(topic)
+      }
     }
 
     // Remove this peer from the mesh
@@ -1191,17 +1194,21 @@ export class GossipSub extends TypedEventEmitter<GossipsubEvents> implements Pub
     this.log('subscription update from %p topic %s', from, topic)
 
     let topicSet = this.topics.get(topic)
-    if (topicSet == null) {
-      topicSet = new Set()
-      this.topics.set(topic, topicSet)
-    }
 
     if (subscribe) {
+      if (topicSet == null) {
+        topicSet = new Set()
+        this.topics.set(topic, topicSet)
+      }
+
       // subscribe peer to new topic
       topicSet.add(from.toString())
-    } else {
+    } else if (topicSet != null) {
       // unsubscribe from existing topic
       topicSet.delete(from.toString())
+      if (topicSet.size === 0) {
+        this.topics.delete(topic)
+      }
     }
 
     // TODO: rust-libp2p has A LOT more logic here
--- a/test/gossip.spec.ts
+++ b/test/gossip.spec.ts
@@ -269,6 +269,84 @@ describe('gossip', () => {
     expect(peerInfoB?.tags.get(topic)).to.be.undefined()
   })
 
+  it('should delete empty topic entries after remote unsubscribe', async function () {
+    this.timeout(10e4)
+    const nodeA = nodes[0]
+    const nodeB = nodes[1]
+    const topic = 'empty-topic-cleanup'
+
+    await connectAllPubSubNodes([nodeA, nodeB])
+
+    nodeA.pubsub.subscribe(topic)
+    await pEvent(nodeB.pubsub, 'subscription-change')
+
+    expect((nodeB.pubsub as any).topics.has(topic)).to.be.true()
+
+    nodeA.pubsub.unsubscribe(topic)
+    await pEvent(nodeB.pubsub, 'subscription-change')
+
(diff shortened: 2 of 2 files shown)


wemeetagain:
@lodekeeper please open this PR against the js-libp2p monorepo. This repo will soon be archived

lodekeeper:
Reopened this against the js-libp2p monorepo here:\n\n- libp2p/js-libp2p#3405\n\nThis is the same empty-topic pruning fix from this PR, ported under `packages/gossipsub` in the monorepo per your request.

